PR #6, opened 2015-04-28
folder structure and new script

acaetano:
**Adapted the folder structure to the available script types after 2.4.0 release;**

**Added a HTTP Basic Authentication bruteforce script**
- Do not forget to create the Base64 payload processor in ZAP to encode the generated payload!


fuzzer_http_processor/README.md
```diff
@@ -0,0 +1,4 @@
+Fuzzer HTTP Processor scripts
```
thc202: The name of the directory should match the one in the templates dir (i.e. "httpfuzzerprocessor"), so ZAP properly loads the scripts.

payload_generator/README.md
```diff
@@ -0,0 +1,4 @@
+Payload Generator scripts
```
thc202: The same comment as before (i.e. the underscore char should be removed).

payload_generator/bruteforce.py
```diff
@@ -0,0 +1,54 @@
...
+# This method is called while hasNext() returns true.
+def next():
+    global count;
+    print('next_count = ' + str(count));
+    payload = count;
+    print('payload = ' + str(payload));
+    count+=1;
+    print('incremented next_count = ' + str(count));
+    print(user+':'+passwd[payload]);
+    return user+':'+passwd[payload];
```
thc202: This could return the pair username/password already Base64 encoded, for example:
```Python
import base64
...

return base64.b64encode(user+':'+passwd[payload]);
```
avoids the need to manually add the payload processor, which is error prone.

payload_processor/README.md
```diff
@@ -0,0 +1,4 @@
+Payload Processor scripts
```
thc202: The same comment as for the other "payload" dir.

script_input_vector/README.md
```diff
@@ -0,0 +1,4 @@
+Script Input Vector scripts
```
thc202: This one is the "variant" dir.

thc202:
Sorry for taking so long.
The review is done, let me know if you have any question/comment.

Thank you!
Best regards.

acaetano:
ok, pull request edited according to review comments

httpfuzzerprocessor/http_status_code_filter.py
```diff
@@ -0,0 +1,18 @@
...
+    pass;
+
+# Called after receiving the fuzzed message from the server
+def processResult(utils, fuzzResult) :
+    global http_status_code;
+
+    if(fuzzResult.getHttpMessage().getResponseHeader().getStatusCode() == http_status_code):
+        return bool(0);
+    
+    # Returns true to accept the result, false to discard and not shown it
```
thc202: There's a typo, "shown" should be "show".
It was fixed in the original file (but not released yet).

thc202:
Sorry once again for the late reply.

New changes look good to me :)

Would you mind squashing the fixes?

acaetano:
Commits squashed and ready for ignition sequence!

thc202:
Don't want to be a PITA with this but, could you split the restructuring changes from the fuzzer processors and payload generator changes? And update the commit message to reflect only the current changes?

acaetano:
Do you mean 4 commits? one for the restructuring changes and another one for each script?

acaetano:
There it is. That's the best i could do, as the final folder structure was committed after the first script...

thc202:
You can change the first commit to match the final folder structure, or is that to keep the history?

acaetano:
I'm really thinking about recreating the repo and the pull request...

thc202:
@acaetano you don't need to do that, you just need to edit the existing commits. Let me know if I can be of help.

acaetano:
I'm sorry, but i never did that. I searched about it but the process seems too much work and very prone to errors that could make it even worse.

PR #9, opened 2015-08-04
Create json_csrf_poc_generator

haseebeqx:
CSRF POC Generator that can also generate csrf forms for websites which use json in requests. it can also work as regular csrf poc generator.


psiinon:
A useful script :)
One suggestion - you could save the form to a local (temporary?) HTML file, and then open it an the browser using DesktopUtils.openUrlInBrowser(url) https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/utils/DesktopUtils.java#L55

PR #10, opened 2015-08-10
json_poc_generator now copies to clipboard

haseebeqx:
made necessary edits to copy the output to clipboard


targeted/json_csrf_poc_generator
```diff
@@ -1,38 +1,43 @@
...
 
 function invokeWith(msg) {
-	print("\n\n\n");
-	print("<!DOCTYPE html>\n");
-	print("<head>\n <title>CSRF POC</title>\n</head>");
-	print("<body>");
+	string = "<!DOCTYPE html>\n";
+	string += "<head>\n <title>CSRF POC</title>\n</head>";
+	string += "\n<body>";
 	if(msg.getRequestHeader().getMethod().toString()== "POST"){
```
thc202: The method ```toString()``` can be removed.

targeted/json_csrf_poc_generator
```diff
@@ -1,38 +1,43 @@
...
 	body = msg.getRequestBody().toString();
 	body = body.trim();
 	if(isJson(body))
-		print("<form action=\""+ msg.getRequestHeader().getURI().toString() +"\" id=\"formid\" method=\"post\" enctype=\"text/plain\">" );
+		string += "\n<form action=\""+ msg.getRequestHeader().getURI().toString() +"\" id=\"formid\" method=\"post\" enctype=\"text/plain\">" ;
 	else
-		print("<form action=\""+ msg.getRequestHeader().getURI().toString() +"\" id=\"formid\" method=\"post\">");
+		string += "\n<form action=\""+ msg.getRequestHeader().getURI().toString() +"\" id=\"formid\" method=\"post\">";
 	if(!isJson(body)){	
 	body=body.split('&');
```
thc202: Might worth checking first that the ```Content-Type``` of the request is ```x-www-form-urlencoded```.

targeted/json_csrf_poc_generator
```diff
@@ -1,38 +1,43 @@
...
 	for(i=0;i<body.length;i++)
 	{
 		keyval = body[i].split('=')
-		print("<input type=\"hidden\" name=\""+decodeURIComponent(keyval[0])+"\" value=\""+decodeURIComponent(keyval[1])+"\" />");	
+		string += "\n<input type=\"hidden\" name=\""+decodeURIComponent(keyval[0])+"\" value=\""+decodeURIComponent(keyval[1])+"\" />";
```
thc202: The name/value of the parameter should be HTML escaped (unless that's on purpose?).

targeted/json_csrf_poc_generator
haseebeqx: i added multipart data support

targeted/json_csrf_poc_generator
haseebeqx: why should i do that. since it's purpose is csrf poc generating from an http request in sites tree. i don't think it is needed.

haseebeqx:
updated again. thanks for your review

targeted/json_csrf_poc_generator
thc202: It's probably better to still check that the content type is the expected one (```x-www-form-urlencoded```), no point continue parsing the body if it's not that type.

targeted/json_csrf_poc_generator
thc202: If the name/value are not escaped it might break the HTML.
For example, if a form field was submitted with ```"/><script>alert(':)')</script>``` (which is not unlikely to happen given the nature of ZAP).

thc202:
Oh, it's missing the file extension.

targeted/json_csrf_poc_generator
haseebeqx: I don't think anyone would want to generate a csrf it if that's the case

targeted/json_csrf_poc_generator
haseebeqx: As the purpose is of a targeted csrf poc generator. I think it's not
necessary

haseebeqx:
fixed it

targeted/json_csrf_poc_generator
thc202: OK.

targeted/json_csrf_poc_generator
thc202: True, that's just an extreme example, but other "valid" (i.e. more normal) values can break the HTML too.

haseebeqx: Whatever the situation is, I don't think that's needed for a csrf poc
generator.
Considering what a normal user needs.

thc202: OK.

targeted/json_csrf_poc_generator.js
```diff
@@ -0,0 +1,97 @@
...
+		string += "\n<form action=\""+ msg.getRequestHeader().getURI().toString() +"\" id=\"formid\" method=\"post\" enctype=\"text/plain\">" ;
+	else{
+		if(ismultipart(msg.getRequestHeader()))
+			string += "\n<form action=\""+ msg.getRequestHeader().getURI().toString() +"\" id=\"formid\" enctype=\"multipart/form-data\" method=\"post\">";
+		else
+			string += "\n<form action=\""+ msg.getRequestHeader().getURI().toString() +"\" id=\"formid\" method=\"post\">";
+	}
+	if(!isJson(body)){	
+	if(ismultipart(msg.getRequestHeader())){
+		type 	=  msg.getRequestHeader().getHeader(org.parosproxy.paros.network.HttpRequestHeader.CONTENT_TYPE);
```
thc202: Probably better to use "HttpHeader" instead of "HttpRequestHeader", the latter does not work in Nashorn.

targeted/json_csrf_poc_generator.js
```diff
@@ -0,0 +1,97 @@
...
+		delim 	=  type.substring(type.search("=")+1,type.length());
+		h = body.split("--"+delim);
+		k=0;
+		names = [];
+		values = [];
+		for(i =1 ; i<h.length-1;i++){
+			j = h[i].split(msg.getRequestHeader().getLineDelimiter());
```
thc202: Use "\r\n" instead (if the header is using unexpected line endings ("\n") it could break the parsing of the body).

targeted/json_csrf_poc_generator.js
```diff
@@ -0,0 +1,97 @@
...
+			nameField =j[1].substring( j[1].search("name")+5,j[1].length());
+			start = nameField.indexOf("\"")+1;
+			end = nameField.indexOf("\"",start);
+			names[k] = nameField.substring(start,end);
+			for(ii=2;ii<j.length-1;ii++){
+				if(j[ii].length() == 0) //find a blank line
+					break;
+			}
+			if(ii == j.length-1)
+				values[k] = "";
```
thc202: This statement can be extract from both if/else blocks (allowing to remove the "else" block).

targeted/json_csrf_poc_generator.js
```diff
@@ -0,0 +1,97 @@
...
+			else
```
thc202: Missing "{" ?

targeted/json_csrf_poc_generator.js
```diff
@@ -0,0 +1,97 @@
...
+				values[k] = "";
+				while(ii < j.length-1){
```
thc202: The break line characters lost during the split should be added back.

targeted/json_csrf_poc_generator.js
```diff
@@ -0,0 +1,97 @@
...
+					values[k] += j[ii+1].trim();
```
thc202: Trimming might change the contents.

thc202:
Sorry for the delay, review done.

targeted/json_csrf_poc_generator.js
haseebeqx: It was created with Oracle nashorn (zap 2.3 I think). Scripts working with
nashhorn always breaks at rhino saying some simple problems like length is
not a function . I tried to make some compatibility, but  when it works on
one  it breaks on the other. So I recreated it with rhino (so it broke at
nashorn) Forgetting about nashorn.

targeted/json_csrf_poc_generator.js
haseebeqx: It sometime just \n . will the getlinedelimeter()  method give the correct
one ?

haseebeqx:
That's not a problem at all. Thanks

targeted/json_csrf_poc_generator.js
haseebeqx: I don't get it. This script can check if it is a valid json. Simply feed
the output back to it. I tried various json. none brakes the script. Please
specify the input at which it breaks

haseebeqx: ok now i get it thats not a problem. it can work without if else since t is while thanks

targeted/json_csrf_poc_generator.js
haseebeqx: thanks. i did so because the last line is always \n. it will be fixed

haseebeqx:
ok made those edits.

targeted/json_csrf_poc_generator.js
thc202: OK, ZAP is targeting Rhino (Java 7) anyway.

targeted/json_csrf_poc_generator.js
thc202: No, unfortunately. That needs to be checked before/during the parsing.

haseebeqx: how can i ?

thc202: I would check which newline chars are after the first multipart delimiter (guess the first is enough to know which chars are actually being used?).
If it's ```\r\n``` split with ```\r\n``` otherwise use ```\n```.

haseebeqx: to my surprise it worked without any change on both \n and \r\n

haseebeqx: you can check it

PR #8, opened 2015-07-31
Updated google code links to new location on Github

jaredestroud:
The links for additional information about ZAP scripts now reflect their present location on the repo's wiki page and not the old Google code page.

(Extremely small change, more of my OCD than anything)

PR #11, opened 2015-08-10
created curl_command_generator.js

haseebeqx:
creates a curl command based on selected request and add it to clipboard

### DIFF
--- a/targeted/curl_command_generator.js
+++ b/targeted/curl_command_generator.js
@@ -1,0 +1,29 @@
+//it will generate and copy curl commands based on the request
+//released under the Apache v2.0 licence.
+//You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+//auther:@haseebeqx
+
+function invokeWith(msg) {
+	string = "curl -i -s -k -X  '"+msg.getRequestHeader().getMethod().toString()+"'  \\\n";
+	header = msg.getRequestHeader().toString();
+	header = header.split('\n');
+	for(i=1;i<header.length;i++){
+		string += " -H '"+header[i].trim()+"' ";
+	}
+	string += " \\\n";
+	body = msg.getRequestBody().toString().trim();
+	if(body.length() != 0){
+		string += "--data-binary $'"+addSlashes(body)+"' \\\n";
+	}
+	string += "'"+msg.getRequestHeader().getURI().toString()+"'";
+	selected = new java.awt.datatransfer.StringSelection(string);
+	clipboard = java.awt.Toolkit.getDefaultToolkit().getSystemClipboard();
+	clipboard.setContents(selected,null);
+	print (string);
+}
+
+function addSlashes(body){
+	var a ={}
+	a[body] = 1;
+	return JSON.stringify(a).slice(2,-4);
+}
--- a/targeted/curl_command_generator.js
+++ b/targeted/curl_command_generator.js
@@ -5,7 +5,7 @@
 
 function invokeWith(msg) {
 	string = "curl -i -s -k -X  '"+msg.getRequestHeader().getMethod()+"'  \\\n";
-	header = msg.getRequestHeader().toString();
+	header = msg.getRequestHeader().getHeadersAsString();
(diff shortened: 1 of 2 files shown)


targeted/curl_command_generator.js
```diff
@@ -0,0 +1,29 @@
+//it will generate and copy curl commands based on the request
+//released under the Apache v2.0 licence.
+//You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+//auther:@haseebeqx
```
thc202: There's a typo "auther".

targeted/curl_command_generator.js
```diff
@@ -0,0 +1,29 @@
...
+
+function invokeWith(msg) {
+	string = "curl -i -s -k -X  '"+msg.getRequestHeader().getMethod().toString()+"'  \\\n";
```
thc202: The call to ```toString()``` can be removed, ```getMethod()``` already returns a ```String```.

targeted/curl_command_generator.js
```diff
@@ -0,0 +1,29 @@
...
+	header = msg.getRequestHeader().toString();
```
thc202: Since the request line is not used, it's better to use ```getHeadersAsString()``` (instead of ```toString()```) which returns just the headers.

targeted/curl_command_generator.js
```diff
@@ -0,0 +1,29 @@
...
+	header = header.split('\n');
```
thc202: Consider using ```msg.getRequestHeader().getLineDelimiter()``` instead of ```'\n'```.

targeted/curl_command_generator.js
```diff
@@ -0,0 +1,29 @@
...
+	for(i=1;i<header.length;i++){
+		string += " -H '"+header[i].trim()+"' ";
+	}
+	string += " \\\n";
+	body = msg.getRequestBody().toString().trim();
```
thc202: Trimming the body might change the meaning of the request.

targeted/curl_command_generator.js
```diff
@@ -0,0 +1,29 @@
+//it will generate and copy curl commands based on the request
```
thc202: Change ```commands``` to ```command```?

targeted/curl_command_generator.js
haseebeqx: since php does this by default i added that line to work with edited requests.
can you explain how it changes the meaning

haseebeqx: Since PHP does this by default i done this for working with edited requests.
can you explain how it might change the meaning

targeted/curl_command_generator.js
haseebeqx: it gives an error
java.lang.reflect.UndeclaredThrowableExceptionsun.org.mozilla.javascript.internal.EcmaError: TypeError: Cannot find function getHeadersAsString in object org.parosproxy.paros.network.HttpMessage@4222af56. (<Unknown source>#8) in <Unknown source> at line number 8

haseebeqx:
updated based on thc202's recommendations.
thanks thc202

targeted/curl_command_generator.js
thc202: Note that it's instead of ```toString()```, that is, ```msg.getRequestHeader().getHeadersAsString()```.
Sorry for not being clear.

targeted/curl_command_generator.js
thc202: Since leading and trailing whitespace chars are removed from the string it changes its contents which most likely will also change its meaning. That can easily happen with "binary" bodies (for example, with AMF message, the ```trim()``` removes the version and header count (and possibly a couple more bytes, I think) which breaks the message).
The function ```addSlashes```might change the meaning of the request, too.

I'm OK with that, although it should be described in the script that not all types of messages are correctly handled, or that it's most appropriate for "text" bodies.

haseebeqx:
that seams to be a good reason. trim removed.

PR #12, opened 2015-08-10
created search cvedetails using target server header.js

snoopythesecuritydog:
scripts takes server header, opens the web browser and conducts a search on cvedetails.com

### DIFF
--- a/targeted/search cvedetails using target server header.js
+++ b/targeted/search cvedetails using target server header.js
@@ -1,0 +1,13 @@
+// Captures Server header from the application response and searches cvedetails.com for known target server vulnerabilities.
+
+
+function invokeWith(msg) {
+	host = msg.getRequestHeader().getURI().getHost(); 
+if (msg) 
+	{
+		headers = msg.getResponseHeader().getHeaders("Server")
+		org.zaproxy.zap.utils.DesktopUtils.openUrlInBrowser(
+		"http://www.cvedetails.com/google-search-results.php?q=" + headers + "&sa=Search");
+	 }
+    
+}


targeted/search cvedetails using target server header.js
```diff
@@ -0,0 +1,13 @@
+// Captures Server header from the application response and searches cvedetails.com for known target server vulnerabilities.
+
+
+function invokeWith(msg) {
+	host = msg.getRequestHeader().getURI().getHost();
```
thc202: This statement should be removed, ```host``` is not used.

targeted/search cvedetails using target server header.js
```diff
@@ -0,0 +1,13 @@
+// Captures Server header from the application response and searches cvedetails.com for known target server vulnerabilities.
+
+
+function invokeWith(msg) {
+	host = msg.getRequestHeader().getURI().getHost(); 
+if (msg)
```
thc202: This check is not needed (```msg``` is always defined, if not it would be a bug in ZAP).

targeted/search cvedetails using target server header.js
```diff
@@ -0,0 +1,13 @@
+// Captures Server header from the application response and searches cvedetails.com for known target server vulnerabilities.
+
+
+function invokeWith(msg) {
+	host = msg.getRequestHeader().getURI().getHost(); 
+if (msg) 
+	{
+		headers = msg.getResponseHeader().getHeaders("Server")
```
thc202: The method ```getHeaders(String)``` returns a list of headers (if you just want one you should use ```getHeader(String)```, which returns a ```String```). If you really want to check for several ```Server``` headers it's probably better to create the query parameter's value manually (i.e. concatenate the values of the list) otherwise the value of ```q``` ends up with chars ```[``` and ```]``` (added because of conversion from list to string).
In both cases the return value is not defined if no ```Server``` header is present.

targeted/search cvedetails using target server header.js
```diff
@@ -0,0 +1,13 @@
...
+		org.zaproxy.zap.utils.DesktopUtils.openUrlInBrowser(
+		"http://www.cvedetails.com/google-search-results.php?q=" + headers + "&sa=Search");
```
thc202: The value of ```headers``` should be URL encoded otherwise it might fail to create a valid URL (or the one intended).

snoopythesecuritydog:
Thanks, will make those changes.

PR #15, opened 2015-08-12
Active Scan Python script to test if the webserver has potentially insecure http methods enabled

gmaran23:
Active Scan Python script to test if the webserver has potentially
insecure http methods enabled
Author: http://renouncedthoughts.wordpress.com

Tested to work with some of the online vulnerable applications like:
http://zero.webappsecurity.com
http://testaspnet.vulnweb.com
http://testfire.net/
http://crackme.cenzic.com

and other apps from
https://www.owasp.org/index.php/OWASP_Vulnerable_Web_Applications_Directory_Project#tab=On-Line_apps


active/TestInsercureHTTPVerbs.py
```diff
@@ -0,0 +1,68 @@
...
+acceptedhttpstatuscodesforinsecureverbs=[301, 302, 404, 405, 403, 400, 501]
+activevulnerabilitytitle='Potentially Insecure HTTP Verb allowed'
+activevulnerabilityfulldescription='Some of the HTTP methods can potentially pose a security risk for a web application, as they allow an attacker to modify the files stored on the web server and, in some scenarios, steal the credentials of legitimate users.' + 'Insecure configuration can possibly lead to web server compromise and website defacement. ' + 'If an application needs one or more of the potentially insecure HTTP methods, such as for REST Web Services (which may require PUT or DELETE), it is important to check that their usage is properly limited to trusted users and safe conditions. ' + 'http://security.stackexchange.com/questions/21413/how-to-exploit-http-methods.'
+activevulnerabilitysolution = 'Configure the web server to allow insecure methods like DELETE and PUT only for the relevant resources. ' + 'If your application does not need HTTP methods other than GET and POST, consider disabling the unused HTTP methods.'
+
+
+def PrepareHTTPRequest(insecureverb, uriparsed):
```
thc202: It's better to use ```sas.sendAndReceive(HttpMessage)``` instead of sending the request "manually". As is now it doesn't use the ZAP configurations (outgoing proxy, authentication...) nor are the requests shown in the "Active Scan" tab.

active/TestInsercureHTTPVerbs.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def GetHTTPResponse(insecureverb, uriparsed):
+    connection = PrepareHTTPRequest(insecureverb, uriparsed)
+    return connection.getresponse();
+
+
+def PrintAlerts(sas, msg, uri, insecureverb, responsestatuscode, responsestatusmessage):
+    attackevidence = 'VERB: \t' + insecureverb + '\t-- STATUS: '+ str(responsestatuscode) +' -- \tMESSAGE: ' + responsestatusmessage
+    print (attackevidence)
```
thc202: Consider adding a flag to control the printing of debug(?) information.

active/TestInsercureHTTPVerbs.py
```diff
@@ -0,0 +1,68 @@
...
+    sas.raiseAlert(1, 2, activevulnerabilitytitle, activevulnerabilityfulldescription, uri, 'HTTP VERB', 'ZAP sent an HTTP request with method - ' + insecureverb, '', activevulnerabilitysolution, attackevidence, 0, 0, msg);
```
thc202: The message raised in the alert should be the one that shows/triggers the vulnerability (not the original one), otherwise it's not easier to evaluate the reported alert.

active/TestInsercureHTTPVerbs.py
```diff
@@ -0,0 +1,68 @@
...
+            httpresponse = GetHTTPResponse(insecureverb, uriparsed)
+            if not httpresponse.status in acceptedhttpstatuscodesforinsecureverbs:
+                PrintAlerts(sas, msg, uri, insecureverb, httpresponse.status, httpresponse.reason)
+        except Exception, e:
+		  print('ERROR For: ' + insecureverb + 'Detail: ' + e.message)
+
+
+def scanNode(sas, msg):
+  url = msg.getRequestHeader().getURI().toString()
+  print('active scan script called for url=' + url + '\n');
```
thc202: Same comment as above regarding the flag to control printing.

thc202:
Could you change the subject of the commit message so that it doesn't show ```"""```?

gmaran23:
excellent review comments. Will refactor and submit my changes.

thc202:
There's a typo in the file name ("Inse**r**cure").

kingthorin:
If you could just do a fix up so this is all one commit, I think we'd be good to go.

kingthorin:
@gmaran23 If you could just do a fix up so this is all one commit, I think we'd be good to go.

thc202:
It would be great if (at least) the comment https://github.com/zaproxy/community-scripts/pull/15#discussion_r37613704 was addressed before merging.

active/TestInsecureHTTPVerbs.py
```diff
@@ -0,0 +1,68 @@
...
+
+and other apps from https://www.owasp.org/index.php/OWASP_Vulnerable_Web_Applications_Directory_Project#tab=On-Line_apps
+
+"""
+
+import httplib
+import urlparse
+
+insecureverbs=["OPTIONS", "HEAD", "PUT", "DELETE", "TRACE", "CONNECT", "DEBUG", "MOVE", "SEARCH", "PATCH", "MKCOL", "COPY", "LOCK", "UNLOCK", "ARBIT", "XXXX", "12AB"]
+acceptedhttpstatuscodesforinsecureverbs=[301, 302, 404, 405, 403, 400, 501]
```
kingthorin: If these are the accepted statuses shouldn't it include `200`?

active/TestInsecureHTTPVerbs.py
```diff
@@ -0,0 +1,68 @@
...
+    print (attackevidence)
+    sas.raiseAlert(1, 2, activevulnerabilitytitle, activevulnerabilityfulldescription, uri, 'HTTP VERB', 'ZAP sent an HTTP request with method - ' + insecureverb, '', activevulnerabilitysolution, attackevidence, 0, 0, msg);
+
+
+def TestTheURIForInsecureVerbs(sas, msg, uri, insecureverbs):
+    uriparsed = urlparse.urlparse(uri)
+    for insecureverb in insecureverbs:
+        try:
+            httpresponse = GetHTTPResponse(insecureverb, uriparsed)
+            if not httpresponse.status in acceptedhttpstatuscodesforinsecureverbs:
```
kingthorin: Not in accepted? (Isn't that backwards?)

kingthorin:
@gmaran23 any feedback?

gmaran23:
@kingthorin 

Imagine that TRACE is an insecure HTTP verb and it should be disabled at the web server. If TRACE is disabled then we would love to see a 401 (unauthorized) or a 405 (not allowed) or a 501 (not implemented). However, I have seen different implementations where if an insecure verb is not allowed, some webservers redirect (301 or 302), some webservers return a 404.

So, if an insecure HTTP verb is not allowed, then the HTTP responses can be 401, 404, 403, 501 .... and those are accepted status codes.

If the HTTP response returns anything other than the accepted status codes (401, 404, 403, 501), then it could potentially mean that the web server has an insecure verb enabled.

Does that help?

kingthorin:
It would probably be good to include your example inside the script so that others can understand the thought process.

Also I think @thc202 had mentioned something that needed fixing before merge. It should also be fixed up into a single commit via rebase interactive.

gmaran23:
Sure, I will include the comments in the script. And I will fix the comments from @thc202 to make HTTP requests via sas.sendAndReceive(HttpMessage) instead of directly invoking the python httplib classes.

I haven't played around with the sas.sendAndReceive(HttpMessage) and what it returns, so I am going to have to experiment and then commit to this pull request.

kingthorin:
Thanks

thc202:
@gmaran23 that method does not return anything, it updates the response (header and body) of the message passed as parameter (if sent successfully). You can then get the status code with `msg.getResponseHeader().getStatusCode()`.

gmaran23:
@thc202 thank you very much. Much appreciated. I will post when when needed help.

kingthorin:
@gmaran23 is there an update coming for this that includes the items discussed above?

gmaran23:
@kingthorin Apologies for the prolonged delay.

Refactored the entire script file using sendAndReceive(HttpMessage) and also added a flag to control printing of debug messages.

Let me know if there's further review comments.

active/TestInsecureHTTPVerbs.py
```diff
@@ -0,0 +1,71 @@
...
+		print('After mutating resulting HTTP VERB: -- \t' + msg.getRequestHeader().getMethod() + '\n')
+
+	
+def TestTheURIForInsecureVerbs(sas, msg, uri, insecureverbs):
+	for insecureverb in insecureverbs:
+		try:
+			if printdebugmessages:
+				print ('Testing Initiated for the HTTP VERB: -- \t' + insecureverb + '\n')
+			PrepareHttpRequest(msg, insecureverb)
+			sas.sendAndReceive(msg)
```
thc202: This should call `sendAndReceive(msg, False)`, to not follow redirects. Sorry, for noticing in the first review.

active/TestInsecureHTTPVerbs.py
```diff
@@ -0,0 +1,71 @@
...
+		except Exception, e:
+			print('ERROR For: ' + insecureverb + 'Detail: ' + e.message + '\n')
+
+
+def scanNode(sas, msg):
+	url = msg.getRequestHeader().getURI().toString()
+	if printdebugmessages:
+    		print('Active scan script called for url = ' + url + '\n');
+		print('Testing for insecure verbs against a list of accepted http response status codes for the url = ' + url + '\n')
+    	msg = msg.cloneRequest();
```
thc202: This should be done before `PrepareHttpRequest` in `TestTheURIForInsecureVerbs` instead of here, we need a new message for each test (otherwise the message shown in the active scan and alert might not be correct).

thc202:
Could you add the following function to the script?
```Python
def scan(sas, msg, param, value):
    pass
```
unfortunately the Python script engine returns always an interface even if the script does not implement it leading to errors in the script (and consequently be disabled).

active/TestInsecureHTTPVerbs.py
gmaran23: this review comment is a lifesaver. I have been bugged by same headers shows in Active Scan tab and Alerts tab.

Thanks @thc202 Good review!

active/TestInsecureHTTPVerbs.py
gmaran23: done.

gmaran23:
@thc202 review comments honored. Thank you. Let me know if anything else is missing.

thc202:
Thanks! Looks good, the commits could be squashed into one but we can also do that on merging.

PR #16, opened 2015-08-13
Community scripts revised versions

snoopythesecuritydog:
Closed the previous pull request to make changes. This pull request has the following scripts

1.passive script to find base64 encoded strings
2. search cvedetails using target server header

### DIFF
--- a/targeted/search cvedetails using target server header.js
+++ b/targeted/search cvedetails using target server header.js
@@ -1,0 +1,11 @@
+// Captures Server header from the application response and searches cvedetails.com for known target server vulnerabilities.
+
+
+function invokeWith(msg) {
+
+		headers = msg.getResponseHeader().getHeader("Server")
+		org.zaproxy.zap.utils.DesktopUtils.openUrlInBrowser(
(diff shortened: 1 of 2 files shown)


targeted/search cvedetails using target server header.js
```diff
@@ -0,0 +1,11 @@
+// Captures Server header from the application response and searches cvedetails.com for known target server vulnerabilities.
+
+
+function invokeWith(msg) {
+
+		headers = msg.getResponseHeader().getHeader("Server")
```
thc202: Consider renaming the variable to ```header```.
It should be checked that the header is defined. If response does not have the ```Server``` header it will be searched for "null".

thc202:
Could you remove the merge commits?

snoopythesecuritydog:
Thanks, will make those changes and start a new pull request

thc202:
Note that you can update this pull request (you just need to leave it open and push the new changes).

snoopythesecuritydog:
Ok, thanks for the info. I am new to GIthub

thc202:
You're welcome! You can change your local master branch and push directly to it, no need to use another branch to introduce new changes (also a rebase interactive should get rid of merge commits).

PR #21, opened 2015-08-26
New active script to provide Good-Old-Files (gof) functionality

kingthorin:
This is a new active script which brings Good-Old-Files (gof) [https://github.com/hacktics/good-old-files] functionality to ZAP. It is a work in progress including a TODO list at the bottom of the file as was discussed here: https://groups.google.com/forum/#!topic/zaproxy-users/DcJrLxYd4x4

This script actively scans by altering the original requested URL in an attempt to find backup or archived versions of web content and components. Such as http://example.com/index.html also being available as: http://example.com/index.html.bak

It's been tested against part of WAVSEP 1.5 as well as other static mockups.

### DIFF
--- a/active/gof_lite.js
+++ b/active/gof_lite.js
@@ -1,0 +1,123 @@
+// The scanNode function will typically be called once for every page 
+// The scan function will typically be called for every parameter in every URL and Form for every page 
+
+// Note that new active scripts will initially be disabled
+// Right click the script in the Scripts tree and select "enable"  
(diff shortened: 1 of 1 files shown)


psiinon:
A bit more detail in the initial comment would be good, but otherwise looks good to me

kingthorin:
Done.

thc202:
Thanks for the scripts!

PR #22, opened 2015-08-27
Added httpsender script to handle anti CSRF token

dmettem:
At the moment ZAP supports only FORM parameter tokens. This script is designed to look for a CSRF token value in the response body using a regex pattern match. Besides capturing the CSRF token value from response body, it also uses the latest value for all outgoing requests irrespective of where the CSRF token is located i.e. FORM parameters or URL parameters or Cookie parameters.

This script is generic, however, it requires enough configuration setup for application under test. Following are several configuration entries that need to be modified in the script.
antiCsrfTokenName - Anti CSRF Token name in the application under test.
urlRegEx - Regex for the request URL that returns CSRF token value in its response body.
csrfTokenValueRegEx - Regex for capturing the CSRF token value from the response body. Make sure to use grouping in the regex, to capture a certain portion of the match.
matcherGroupNumber - Group number to be matched and captured as new Anti CSRF token value. This value will be replaced in all subsequent requests that has a CSRF token in the request.
parameterTypesList - An array of one or more parameter types to be looked in, for replacing the anti CSRF token value in the outgoing requests. Supported values in the array: formParamType, urlParamType, cookieParamType.

### DIFF
--- a/httpsender/Capture and Replace Anti CSRF Token.js
+++ b/httpsender/Capture and Replace Anti CSRF Token.js
@@ -1,0 +1,95 @@
+// At the moment ZAP supports only FORM parameter tokens. This script is designed to look for a CSRF token value in the response body using a regex pattern match. Besides capturing the CSRF token value from response body, it also uses the latest value for all outgoing requests irrespective of where the CSRF token is located i.e. FORM parameters or URL parameters or Cookie parameters.
+// This script is generic, however, it requires enough configuration setup for application under test.
+// Following are several configuration entries that need to be modified in the script.
+// antiCsrfTokenName - Anti CSRF Token name in the application under test.
+// urlRegEx - Regex for the request URL that returns CSRF token value in its response body.
+// csrfTokenValueRegEx - Regex for capturing the CSRF token value from the response body. Make sure to use grouping in the regex, to capture a certain portion of the match.
+// matcherGroupNumber - Group number to be matched and captured as new Anti CSRF token value. This value will be replaced in all subsequent requests that has a CSRF token in the request.
+// parameterTypesList - An array of one or more parameter types to be looked in, for replacing the anti CSRF token value in the outgoing requests. Supported values in the array: formParamType, urlParamType, cookieParamType.
+
+// REPLACE the values for the variables as applicable to your application.
+
+// Regular expression for the request URI that returns CSRF token in response.
+// If the application under test returns csrf token in every response or in response to more than request, set a generic regex that matches with host name or domain name of the application.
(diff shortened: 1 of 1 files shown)


httpsender/Capture and Replace Anti CSRF Token.js
```diff
@@ -0,0 +1,101 @@
...
+
+function modifyParams(parameterType, params) {
+    var iterator = params.iterator();
+    while(iterator.hasNext()) {
+        var param = iterator.next();
+        // Check if the url parameters has the antiCsrfTokenName in it.
+        if (param.getName().equals(antiCsrfTokenName)) {
+            var secureTokenValue = param.getValue();
+            // Check for the value of AntiCsrfTokenName in the existing request with the latest value captured from previous requests.
+            if (!secureTokenValue.equals(org.zaproxy.zap.extension.script.ScriptVars.getGlobalVar("anti.csrf.token.value"))) {
```
thc202: Might worth checking that the variable is defined before checking that's equal.
It's not defined in the following case:
 1. User accesses the page that submits the anti CSRF token;
 1. User adds and configures the script;
 1. User submits the request;
 1. Token value is replaced with ```null```.

httpsender/Capture and Replace Anti CSRF Token.js
```diff
@@ -0,0 +1,101 @@
...
+                // Remove the antiCsrfToken from the form parameters.
+                var oldCsrfToken = new org.parosproxy.paros.network.HtmlParameter(parameterType, antiCsrfTokenName, secureTokenValue);
+                params.remove(oldCsrfToken);
```
thc202: This and the above statement could be replaced with ```iterator.remove();```.
Or, since it's just being changed the value of the parameter the above and following two statements could be replaced with:
```JavaScript
param.setValue(org.zaproxy.zap.extension.script.ScriptVars.getGlobalVar("anti.csrf.token.value"))
```
It also makes the parameter ```parameterType``` obsolete.

httpsender/Capture and Replace Anti CSRF Token.js
dmettem: I think, it is. Change will be part of my next commit.

httpsender/Capture and Replace Anti CSRF Token.js
dmettem: That's a good point. Made the change. Next commit will include this too.

dmettem:
Review comments are fixed.

thc202:
Thank you!

PR #24, opened 2015-09-01
updated and renamed csrf_poc_generator for json

haseebeqx:
no need to check for /r/n it works in both cases for the  intended purpose  of a csrf poc generator.  i checked with multiple requests and verified it is working without edits.
so please merge this one


kingthorin:
Looks fine to me.

haseebeqx:
it works fine under expected scenarios. i don't see any reasons to add bit level checking for simple csrf generator which generates csrf html page as a form tag in a web page.  i far as i see this one is more advanced than many other csrf generaors. and also even built in one dont have these bit level checkings. because that is absolutely not needed. it is a big disappointment that this one is not merged yet.

thc202:
For the record #10.

PR #18, opened 2015-08-19
added body to xml converter script

haseebeqx:
this one can act as an assistant in finding xxe.
it supports normal body, multipart body and json and convert them to xml
then it open manual request editor with this data.
further editing is required if it has to be useful for finding vulnerabilities  
it will also change the content type to application/xml

PR #25, opened 2015-09-01
added request to xml converter script

haseebeqx:
made the requested edits


haseebeqx:
why not yet merged. i tested it with windows and linux. it is working

thc202:
For the record #18.

PR #27, opened 2015-09-23
Update gof_lite.js add check and handling for users stopping scans

kingthorin:
Now checks and handles `this.isStop()`, so that users can choose to stop their scan while this active script is running.

### DIFF
--- a/active/gof_lite.js
+++ b/active/gof_lite.js
@@ -16,6 +16,7 @@
 // gof_lite.js
 // Author: kingthorin+owaspzap@gmail.com
 // 20150828 - Initial submission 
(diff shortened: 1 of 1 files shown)


active/gof_lite.js
```diff
@@ -51,6 +52,9 @@ function scanNode(as, msg) {
 	}
 
 	for(idx in mutationStrings) {
+		if(this.isStop()) { //Check if the user stopped the scan
```
thc202: The `this` should be `as` (i.e. `ScriptsActiveScanner`).

kingthorin:
Thanks, will try that out and submit another update.

PR #31, opened 2015-11-20
Add script to handle CAS authentication

tfcporciuncula:
As it was suggested in [this topic](https://groups.google.com/forum/#!topic/zaproxy-scripts/a7Tv8VE5C4Q) at the [OWASP ZAP Scripts group](https://groups.google.com/forum/#!forum/zaproxy-scripts), I'm submitting the CAS authentication script.


kingthorin:
I haven't tested this, but I have read through it and it seems ok to me.

PR #32, opened 2016-01-13
RDBMS connection string key word passive scan

bowserjklol:
Hey, been using ZAP a lot recently and figured I should contribute.  I've been finding different database connection strings stored in client-side JS functions on two recent pen tests.  Figured if I've found it twice, this (bad and unnecessary) practice is likely more wide spread.  The new passive script will regexp compare the response body with various connection string key words, alerting a high alert if found.  Hope this helps someone...


kingthorin:
Thanks for the contribution.

Any chance you could do an interactive rebase and make this all into one commit with fixups? [1][2]

Also do you have or could you mockup an example of what you encountered that inspired this? (You could set it up as a public Gist, give us a pastebin link, point us to a blog article, whatever....I'm just curious and a workable testcase is never a bad thing. Or if it's just not possible I can get that too.)

[1] https://wiki.eclipse.org/EGit/User_Guide#Interactive_Rebase
[2] https://www.kernel.org/pub/software/scm/git/docs/git-rebase.html#_interactive_mode

passive/Find Data Store Credentials.js
```diff
@@ -0,0 +1,98 @@
...
+
+	//create an array for matching strings as we might have more than 1 hit
+	var credsFound = []
+
+	//now for loop through the array
+	for (var i=0; i < dataSourceCheck.length; i++)
+	{
+
+		//If test the body against our regexp array
+		if(dataSourceCheck[i].test(body)) //test() returns true/false boolean so we know to continue
```
thc202: Is this check (and following statement) really needed? It could just do the exec(s). The code is not harder to read without it (it's thoroughly commented anyway).

passive/Find Data Store Credentials.js
```diff
@@ -0,0 +1,98 @@
...
+	if (credsFound.length != 0)
+	{
+
+		//finally, we have our array of hits, raise the alert
+		ps.raiseAlert(alertRisk
+			, alertReliability
+			, alertTitle
+			, alertDesc
+			, url
+			, 'Parm Not Used'
```
thc202: This field should be left empty if it does not apply to the raised alert.

passive/Find Data Store Credentials.js
```diff
@@ -0,0 +1,98 @@
...
+			, 'Attack Not Used'
```
thc202: Same comment as above.

passive/Find Data Store Credentials.js
```diff
@@ -0,0 +1,98 @@
...
+			, credsFound.toString() //these are the matching db credential strings
```
thc202: It would improve readability if the strings were split with a newline.
For example:
```
uid=e.guid=e.guid||x.guid++,i):t},access:function(e,n,r,i,o,a,s){var l=0,u=e.length,c=null==r;,uid=x.guid++),(l=v.events)||(l=v.events={}),
```
It's not obvious where ends a match and starts the other (it requires a little of scanning to spot the semicolon).

passive/Find Data Store Credentials.js
```diff
@@ -0,0 +1,98 @@
...
+			, alertSolution
+			, 'See Other Info Section'
```
thc202: This should/could use the first matched string (ZAP would highlight it, making easier to inspect it in its context).
Otherwise it should be left empty(?).

passive/Find Data Store Credentials.js
bowserjklol: Yep.  I think it makes sense to remove it.  No need to test the body, return true, then immediately exec the body.

passive/Find Data Store Credentials.js
bowserjklol: Yes.  Agreed.  Will update.

passive/Find Data Store Credentials.js
bowserjklol: Good call.  Will change.

bowserjklol:
Seriously, thanks for the review and comments.  

I am definitely new to Git so it might take me screwing around a bit to get the rebase to properly squash these into one commit.  If I can't get it down to one commit with a rebase, what's the best way to move forward?  Kill this and initiate a new pull where I've only got the 1 commit (plus incorporation of the changes you've pointed out)?

As far as a test case for this, I can certainly get something up on gist.  The inspiration for this was two separate pen tests for two different development teams where I found *both* teams had inexplicably pumped their database connection strings into client side javascript functions.  These were then just passed right back to the server in subsequent POSTs.  Seeing it once?  Anomaly.  Twice, a trend.  Had to make a script to check for it.

bowserjklol:
Committed the code changes you suggested.  Have this sitting on gist as a simple test bed for the script. 

https://gist.github.com/bowserjklol/1922ecfbe2a53b79e0e0#file-zapdbcredstest-html

bowserjklol:
ok i just jacked this branch's history up trying to rebase.  I tried to squash just my commits but looks like I've added some other commits to the history inadvertently (excessive use of --skip when I saw errors i guess?).

thc202:
Was this closed on purpose? I was going to suggest a quick fix.

bowserjklol:
I'm open for any suggestion (code change, rebase or other).  I figured it was easier for me to resubmit a pull request with a single commit + the code review comment after hosing up the rebase.

PR #33, opened 2016-01-18
Added IPv6 Variants

bowserjklol:
Was in the to-do list.  Figured I'd *try* and help out with existing scripts if needed.

modified and renamed isIPv4Local function to handle both IPv4 and IPv6

modified decodeIP to handle both IPv4 and IPv6

minor string formatting changes to scan

### DIFF
--- a/passive/f5_bigip_cookie_internal_ip.js
+++ b/passive/f5_bigip_cookie_internal_ip.js
@@ -9,13 +9,16 @@
 // Ref: https://support.f5.com/kb/en-us/solutions/public/6000/900/sol6917.html
 // Author: kingthorin+owaspzap@gmail.com
 // 20150828 - Initial submission
+// 20160117 - Updated to include ipv6 variants - jkbowser[at]gmail[dot]com
 
 function scan(ps, msg, src) {
 	//Setup some details we will need for alerts later if we find something
-	alertRisk = 1
+	alertRisk = [1, 0]
 	alertConfidence = 3
-	alertTitle = 'Internal IP Exposed via F5 BigIP Persistence Cookie'
-	alertDesc = 'The F5 Big-IP Persistence cookie set for this website can be decoded to a specific internal IP and port. An attacker may leverage this information to conduct Social Engineering attacks or other exploits.'
+	alertTitle = ['Internal IP Exposed via F5 BigIP Persistence Cookie'
+				, 'IP Exposed via F5 BigIP Presistence Cookie']
+	alertDesc = ['The F5 Big-IP Persistence cookie set for this website can be decoded to a specific internal IP and port. An attacker may leverage this information to conduct Social Engineering attacks or other exploits.'
+				,'The F5 Big-IP Persistence cookie set for this website can be decoded to a specific IP and port. An attacker may leverage this information to conduct Social Engineering attacks or other exploits.']
 	alertSolution = 'Configure BIG-IP cookie encryption.'
 	alertRefs = 'https://support.f5.com/kb/en-us/solutions/public/6000/900/sol6917.html'
 	cweId = 311
@@ -33,7 +36,7 @@ function scan(ps, msg, src) {
 			cookieValue=cookiesArr[idx].getValue();
 			if(cookieName.toLowerCase().contains("bigip") &&
 			  !cookieValue.toLowerCase().contains("deleted")) {
-				cookieChunks = cookieValue.split("\\."); //i.e.: 3860990474.36895.0000
+				cookieChunks = cookieValue.split(getDelim()); //i.e.: 3860990474.36895.0000
 				//Decode IP
 				try {
 					theIP=decodeIP(cookieChunks[0]);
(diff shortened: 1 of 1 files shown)


kingthorin:
Couple small things:
The existing alert details cover IPv4 for RFC1918 so an alternate title and desc should be used for IPv6 findings.

Could you re-add a todo note at the bottom about IPv6 for non-default route domains? (or tackle that variant?) Per the doc there are two variants for each IP stack [strike through addressed at this point]:
* ~~IPv4 pool members~~
* IPv4 pool members in non-default route domains
* ~~IPv6 pool members~~
* IPv6 pool members in non-default route domains

bowserjklol:
For the existing alert details, they seem open-ended, not specifically mentioning IPv4 directly (below).  Did I over look the specific line you're talking about?  Let me know and I can certainly add in alternate title/desc for IPv6 but they seem ok to me as-is.

```javascript

alertTitle = 'Internal IP Exposed via F5 BigIP Persistence Cookie'
alertDesc = 'The F5 Big-IP Persistence cookie set for this website can be decoded to a specific internal IP and port. An attacker may leverage this information to conduct Social Engineering attacks or other exploits.'

```

kingthorin:
The IPv4 finding is specific to RFC 1918, in fact we don't alert if it isn't internal. 

When I added the todo for IPv6 I only knew that there were such variants I hadn't put any thought at all into the fact that there isn't a IPv6 equivalent to RFC1918....actually now that I'm writing this I think there actually is, but I have no idea how significant it is.

Should we raise 1918s (or equivalent) for both stacks as Low risk, and make externals Informational? I guess that might be a good move, because perhaps someone might leak an external IP that isn't part of their DNS resolution...might indicate load balancing or facilitate Social Engineering of some sort. So less risk but still worth mentioning....food for thought. I'm open to anyone's thoughts on the matter.

Specific to the name I was thinking "Internal IP" should be different for IPv6...potentially.
Specific to the desc I was thinking "....a specific internal IP and port." should probably be worded differently.

bowserjklol:
AH!  Got it.  You're right.  This (and the alertDesc) isn't (always) going to be correct when alerting IPv6.

```javascript
alertTitle = 'Internal IP Exposed via F5 BigIP Persistence Cookie'
```

When working on this try/catch block:

```javascript
try {
			if(java.net.Inet6Address.getByName(ip)) //isSiteLocalAddress - available but not applicable for IPv6 re: http://www.ietf.org/rfc/rfc3879.txt.  Just validating format.
				return true //it is ipv6, may or may not be local
		} catch (e) {
			return false //not ipv6
		}
```

I ended up removing the isSiteLocalAddress call because it only returns TRUE on a single IPv6 address (FEC0).  

https://books.google.co.in/books?id=NyxObrhTv5oC&lpg=PT187&dq=InetAddress.isSiteLocalAddress()&pg=PT186#v=onepage&q&f=false

Then, of course, I find that the direction is to deprecate IPv6 local links all together:

http://www.ietf.org/rfc/rfc3879.txt

This led me to leave it open ended as it may or may not be local since there is no longer a well defined local link range.  At least, that's how I read things.

With all of that in mind, would it make sense for me to make another update where we:

1) Use an array on alertRisk, alertTitle, alertDesc and alertConfidence declarations.
2) For IPv4, existing alertTitle/Desc/Conf stays as-is (or whatever wording makes sense)
3) For IPv6, add back in isSiteLocalAddress and test IPv6 against it.  If true, return the existing IPv4 alert details (since we're confident it's local and reference 0th index in new arrays).  

If it returns false, IF test format with getByName() just to be sure, then return alertDesc[1], etc with new verbiage:

```javascript
alertRisk = 0 //informational
alertTitle = 'IPv6 Address Exposed via F5 BigIP Persistence Cookie'
alertDesc = 'The F5 Big-IP Persistence cookie set for this website can be decoded to a specific IP and port. An attacker may leverage this information to conduct Social Engineering attacks or other exploits.  This address may or may not be internal to the network as it is not an FEC0 address.'
```

Open to any suggestions on a path forward.

kingthorin:
I've got to put some thought into this. For the time being I'm just going to paste one link here and a reference I found: http://tools.ietf.org/html/rfc4193 `fc00::/7`

I'll try to read through your reply and the RFC over the next few days and provide some feedback. (If you haven't heard from me (or someone) by Mon then hit us up with another reply :smile: )

kingthorin:
Ok finally had time to digest this. Your 4 points seem on the right track to me, for #3 I think we should go with the rfc4193 definition of Local IPv6 Unicast Addresses (unless someone else has read something that I've totally missed here).

bowserjklol:
Sounds good to me.  I'll go work on the updates...

kingthorin:
@bowserjklol any news?

bowserjklol:
Wow, it's been a month already.  Will get these changes made in the next few days.

bowserjklol:
Just pushed the latest changes after testing a bit.  Let me know what you think and how you'd like to move forward.

kingthorin:
Ack...will review tonight.

kingthorin:
Wow ok it took me quite a while to get through this. 

It seems that Rhino and Nashorn behave differently as far as split goes. So the split calls should be replaced with calls to getDelim() defined as:
```
function getDelim() {
    //It seems Rhino and Nashhorn behave differently for splitting on period
    if (java.lang.System.getProperty("java.version").startsWith("1.8")) {
        return "\.";
    }
    return "\\.";
}
```

kingthorin:
Thanks @bowserjklol 

I'd like input from people as to whether or not we should be checking both requests and responses?

It seems to be that checking Cookie headers (on request) can lead to a bunch of false positives when cookies are inappropriately scoped (which is often the case).

kingthorin:
Could you do an interactive rebase and fixup the commits?

bowserjklol:
Thanks for your help along the way @kingthorin.  Rebased/fixup is done.  Thanks!

kingthorin:
Great, looks good to me.

PR #34, opened 2016-02-08
New History Source Tagger

kingthorin:
New script that loops through the history table adding source tags in
the form SRC_Proxied, SRC_Manual, SRC_Other. Only tags history items
that don't already contain such a tag. Uses global vars to prevent
repetitive tagging during long sessions.

Addresses zaproxy/zaproxy#78

### DIFF
--- a/standalone/historySourceTagger.js
+++ b/standalone/historySourceTagger.js
@@ -1,0 +1,43 @@
+// Running this script will cause ZAP to loop through all the entries in the history table.
+// Entries in the table will be tagged (in the "tags" column) in the form:
+// SRC_Proxied, SRC_Manual, SRC_Other
+// The script can be run multiple times, history entries will only be tagged
+// if they don't already have a tag that starts with TAG_PREFIX as defined below.
+// Author: kingthorin+owaspzap@gmail.com
+// 20160207: Initial release
+
+extHist = org.parosproxy.paros.control.Control.getSingleton().
+      getExtensionLoader().getExtension(org.parosproxy.paros.extension.history.ExtensionHistory.NAME);
+
+TAG_PREFIX='SRC_';
+
+if (extHist != null) {
+  i=org.zaproxy.zap.extension.script.ScriptVars.getGlobalVar("tagged_ref");// Check for global reference
+  if(i==null) {
+    i=1;// Global reference was null so 1
+  }
+  lastRef=extHist.getLastHistoryId();// Get current max history reference 
+  while (i <= lastRef) {// Loop through the history table
+    hr = extHist.getHistoryReference(i);
+    if(i % 10 == 0 | i == 1 | i == lastRef) {// Progress update first, every 10, and last
+      print('Checking ' + i);
+    }
+    if (hr) { 
+      switch (hr.getHistoryType()) {
+        case 1: type="Proxied"; break;
+        case 15: type="Manual"; break;
+        default: type="Other"; break;
+      }
+      newTag=TAG_PREFIX+type;
+      theTags=hr.getTags();
+      if(!theTags.contains(newTag)) {
+        hr.addTag(newTag);
+        try {
+          extHist.notifyHistoryItemChanged(hr);// Trigger GUI update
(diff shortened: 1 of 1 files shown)


standalone/hsitorySourceTagger.js
```diff
@@ -0,0 +1,48 @@
+// Running this script will cause ZAP to loop through all the entries in the history table.
+// Entries in the table will be tagged (in the "tags" column) in the form:
+// SRC_Proxied, SRC_Manual, SRC_Other
+// The script can be run multiple times, history entries will only be tagged
+// if they don't not already have a tag that starts with TAG_PREFIX as defined below.
```
thc202: The "not" should be removed?

standalone/hsitorySourceTagger.js
```diff
@@ -0,0 +1,48 @@
...
+
+if (extHist != null) {
+  i=org.zaproxy.zap.extension.script.ScriptVars.getGlobalVar("tagged_ref");// Check for global reference
+  if(i==null) {
+    i=1;// Global reference was null so 1
+  }
+  lastRef=extHist.getLastHistoryId();// Get current max history reference 
+  while (i <= lastRef) {// Loop through the history table
+    hr = extHist.getHistoryReference(i);
+      if(i % 10 == 0 | i == 1 | i == lastRef) {// Progress update first, every 10, and last
```
thc202: The indentation in these statements has too much spaces, compared to above statements.

standalone/hsitorySourceTagger.js
```diff
@@ -0,0 +1,48 @@
...
+        print('Checking ' + i);
+      }
+      if (hr) { 
+        switch (hr.getHistoryType()) {
+          case 1: type="Proxied"; break;
+          case 15: type="Manual"; break;
+          default: type="Other"; break;
+        }
+        previouslySet=false;
+        theTags=hr.getTags();
```
thc202: It could use the method contains of the collection to check if the tag was already (or not) added (it would require creating the tag before hand but does not seem to be an issue). The variable "previouslySet" could be removed if the tag is immediately added (if not already).

standalone/hsitorySourceTagger.js
```diff
@@ -0,0 +1,48 @@
...
+            previouslySet=true; // Tag seems to exist
+          }
+        }
+      if(!previouslySet) {
+        hr.addTag(TAG_PREFIX+type);// No previous tag, so add tag
+      }
+    }
+    i++;
+    try {
+      extHist.notifyHistoryItemChanged(hr);// Trigger GUI update
```
thc202: This just needs to be done if the tag was really added, maybe move to if-then block that adds the tag?

standalone/hsitorySourceTagger.js
```diff
@@ -0,0 +1,48 @@
...
+    } catch (ex) { } //Ignore 
+    org.zaproxy.zap.extension.script.ScriptVars.setGlobalVar("tagged_ref",i);// Set global reference
```
thc202: This statement could moved to outside the while block (and could use `lastRef + 1` instead of `i`).

kingthorin:
@thc202 done, fixed the file name too :smiley:

standalone/historySourceTagger.js
```diff
@@ -0,0 +1,43 @@
...
+      if(!theTags.contains(newTag)) {
+        hr.addTag(newTag);
+        try {
+          extHist.notifyHistoryItemChanged(hr);// Trigger GUI update
+        } catch (ex) { } //Ignore 
+      }
+    }
+  i++;
+  }
+org.zaproxy.zap.extension.script.ScriptVars.setGlobalVar("tagged_ref",last_Ref+1); // Set global reference
```
thc202: - There's a typo in `lastRef`;
 - The statement could be indented a little.

thc202:
Nice catch with the name :)

kingthorin:
:zap:

thc202:
Looks good to me. Thank you!

PR #35, opened 2016-02-08
New Domain Finder

kingthorin:
This script performs forward lookups by brute force against a domain of
the user's choice, based on those results it then uses the identified
IPs to perform reverse lookups.

Partially addresses zaproxy/zaproxy#562

### DIFF
--- a/standalone/domainFinder.js
+++ b/standalone/domainFinder.js
@@ -1,0 +1,119 @@
+//REQUIRES: Java 8+
+//
+//Inspired partially by fierce domain scanner
+//Partially addresses https://github.com/zaproxy/zaproxy/issues/562
+//Author: kingthorin+owaspzap@gmail.com
+//20160207: Initial release.
+
+var DOMAIN=".example.org"; //Update this with the domain you want to do lookups on 
+
+var System = Java.type("java.lang.System");
+var Thread = Java.type("java.lang.Thread");
+var TimeUnit = Java.type("java.util.concurrent.TimeUnit");
+var ForkJoinPool = Java.type("java.util.concurrent.ForkJoinPool");
+var ForkJoinTask = Java.type("java.util.concurrent.ForkJoinTask");
+var RecursiveAction = Java.extend(Java.type("java.util.concurrent.RecursiveAction"));
+
+//Prefixes based on http://ftp.isc.org/www/survey/reports/2015/01/first.txt
+var prefixes = ["mail","www","ns2","ns1","server","smtp","mail2","gw","remote","ftp",
+"host","ns","mail1","webmail","mx","mx1","ip1","cpe","vpn","router","mx2","gateway",
+"web","exchange","lo0","server1","vps","mail3","secure","test","ns3","ip2","www2","email",
+"mailhost","dev","dns1","host2","dns2","fw","static","broadcast","host1","eth0","o1","dns",
+"db","net","portal","office","smtp2","e0","owa","proxy","network","admin","lwdc","mta",
+"mail4","host3","adsl","pc1","bcast","web1","se400","mailgate","smtp1","gate","pc2","a",
+"pc3","host4","ns4","pc4","pc5","server2","support","mx3","host5","relay","www1","pc6","e1",
+"nmd","a1","stats","bc","backup","host6","b","sdtc","a0","ip3","mail01","a3","news","c1",
+"a7","b1","firewall"];
+
+var NUMBER_ELEMENTS = prefixes.length;
+// Slice for each RecursiveAction
+var ARRAY_SLICE = NUMBER_ELEMENTS / 2;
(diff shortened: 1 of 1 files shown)


standalone/domainFinder.js
```diff
@@ -0,0 +1,119 @@
+//REQUIRES: Java 8+
+//
+//Inspired partially by fierce domain scanner
+//Partially addresses https://github.com/zaproxy/zaproxy/issues/562
+//Author: kingthorin+owaspzap@gmail.com
+//20160207: Initial relase.
```
thc202: There's a typo in 'release'.

standalone/domainFinder.js
```diff
@@ -0,0 +1,119 @@
...
+var ARRAY_SLICE = NUMBER_ELEMENTS / 2;
+var pool = new ForkJoinPool();
+
+var foundFwd = [];
+var foundRev = [];
+var found_ips = [];
+// Do forward lookups (brute force names to IPs)
+var fb = createRecursiveAction(prefixes, 0, prefixes.length, 'fwd');
+start = startTime();
+pool.invoke(fb);
```
thc202: It could print a message before starting (it might take a while to print the first results).

standalone/domainFinder.js
```diff
@@ -0,0 +1,119 @@
...
+print('');
+printElapsed(start);
+print('Found: ' + foundFwd.length + ' via forward lookup.');
+print('');
+// Do reverse lookups (using the IPs from thw FWDs check if there are other names)
```
thc202: There's a typo `thw`.

standalone/domainFinder.js
```diff
@@ -0,0 +1,119 @@
...
+var fb2 = createRecursiveAction(found_ips, 0, found_ips.length, 'rev');
```
thc202: It could use the same variables from the previous check (at the very least it should (re)use the same pool).

standalone/domainFinder.js
```diff
@@ -0,0 +1,119 @@
...
+        if (host!=undefined & !/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$/.test(host.toString())) {
+          printHost(host);
+          foundRev.push(host);
+        }
+      } catch (e) { return } 
+    case 'fwd':
+      try { 
+        host = java.net.InetAddress.getByName(lookupItem + DOMAIN);
+        if (host!=undefined) {
+          foundFwd.push(host.toString().split('/')[0].trim());
```
thc202: Could be used `host.getHostName()` instead (though this method might return null).

standalone/domainFinder.js
```diff
@@ -0,0 +1,119 @@
...
+          printHost(host);
+          new_ip=host.toString().split('/')[1].trim();
```
thc202: Could be used `host.getHostAddress()` instead.

kingthorin:
Done, thanks for the review.

standalone/domainFinder.js
```diff
@@ -0,0 +1,120 @@
...
+print(foundRev.toString().replaceAll(',', '\n'));
+print('\nThe list above may include CDNs, shared hosts, or 3rd party hosting. Please be careful how you proceed.');
+
+function nslookup(lookupItem, type) {
+  var host='';
+  switch (type) {
+    case 'rev':
+      try {
+        host = java.net.InetAddress.getByName(lookupItem).getCanonicalHostName();
+        if (host!=undefined & !/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$/.test(host.toString())) {
```
thc202: - Noticed now that it's using the bitwise operator (the undefined check could be removed though, from the JavaDoc of the method called it does not seem to return null).
- The `host.toString()` could be replaced with just `host` (it's already a `String`).

standalone/domainFinder.js
```diff
@@ -0,0 +1,120 @@
...
+          printHost(host);
+          foundRev.push(host);
+        }
+      } catch (e) { return } 
+    case 'fwd':
```
thc202: It should `break` before this case.

standalone/domainFinder.js
```diff
@@ -0,0 +1,120 @@
...
+      try { 
+        host = java.net.InetAddress.getByName(lookupItem + DOMAIN);
+        if (host!=undefined) {
```
thc202: This undefined check could also be removed.

kingthorin:
Thanks @thc202

thc202:
It's still missing the `break` statement in "rev" case.

kingthorin:
DOH! Bad copy/paste :disappointed:

kingthorin:
Finally done `<fingers_crossed>`

thc202:
Looks good to me, thank you!

PR #40, opened 2016-06-09
Add two step authentication scripts, for Nashorn and Rhino

aine-rb:
This commit contains two scripts.
The first one can be interpreted by the Nashorn engine, for people executing ZAP in a Java 8 environment for example. The second one can be interpreted by the Rhino engine, for a Java 7 (or less) environment.
They both implement 2-step authentication mechanism: Firstly it sends a POST request with no redirect (because cookies are not set correctly when following a redirect currently), secondly it sends a GET request.
Moreover, a comment showing how to manually set cookies is added in both scripts (this can be useful in some specific cases)

### DIFF
--- a/authentication/NashornTwoStepAuthentication.js
+++ b/authentication/NashornTwoStepAuthentication.js
@@ -1,0 +1,109 @@
+// Author : aine-rb from Sopra Steria (based on the script of thc202 from the OWASP ZAP development team)
+
+// This script is heavily based on the "Simple Form-Based Authentication.js" template
+// It should be interpreted by the Nashorn engine, since Java 8
+// It can be used to authenticate in a webapplication via a form submission followed by a GET request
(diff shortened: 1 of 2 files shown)


kingthorin:
@aine-rb does this request two separate scripts just because of print/println handling?

If so could you just add the helper method that was used [here](https://github.com/zaproxy/community-scripts/pull/52)[1], and reduce this to a single script?

[1] https://github.com/zaproxy/community-scripts/pull/52

thc202:
The Rhino version could be dropped, next main release will require Java 8.

aine-rb:
@kingthorin There is also a difference in the way Java classes are imported between both engines.
@thc202 I added both scripts for people that still use older version of ZAP/Java. I allow edits from maintainers if you want to keep only the Nashorn script.

thc202:
I'm fine adding both, thanks!

PR #45, opened 2016-10-10
Fix broken links in README of Mass Baseline scan

thc202:
Change to use relative links, which requires just the name of the file
"mass-baseline.sh" (the links were leading to 404 as it was missing
"/blob/master/" in the URLs).


kingthorin:
Looks good

PR #48, opened 2016-11-02
In JS,we should use print() function，not println() function

shengulong:
There is no println() function in JS

### DIFF
--- a/standalone/Active scan rule list.js
+++ b/standalone/Active scan rule list.js
@@ -10,16 +10,16 @@ println('\n');
 
 for (var i=0; i < plugins.length; i++) {
   try {
-    println ('Plugin ID: ' + plugins[i].getId());
-    println ('Name: ' + plugins[i].getName());
-    println ('Desc: ' + plugins[i].getDescription());
-    println ('Risk: ' + plugins[i].getRisk());
-    println ('Soln: ' + plugins[i].getSolution());
-    println ('Ref:  ' + plugins[i].getReference());
-    println ('CWE:  ' + plugins[i].getCweId());
-    println ('WASC:  ' + plugins[i].getWascId());
-    println ('');
+    print ('Plugin ID: ' + plugins[i].getId());
+    print ('Name: ' + plugins[i].getName());
+    print ('Desc: ' + plugins[i].getDescription());
+    print ('Risk: ' + plugins[i].getRisk());
+    print ('Soln: ' + plugins[i].getSolution());
+    print ('Ref:  ' + plugins[i].getReference());
+    print ('CWE:  ' + plugins[i].getCweId());
+    print ('WASC:  ' + plugins[i].getWascId());
+    print ('');
   } catch (e) {
-    println (e);
+    print (e);
   }
 }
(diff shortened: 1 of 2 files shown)


kingthorin:
Actually it depends on the engine in use. Java 7 (Rhino) Should work fine Java 8 (Nashorn) will probably not like it.

psiinon:
Yeah. Could we detect if print is available and if not create that function? Thats the sort of thing I've seen done in other places.

kingthorin:
Could do something like this:
```JavaScript
eval("if (typeof println == 'undefined') this.println = print;");

println("Fred");
print("Bob");
```

psiinon:
That sounds good - maybe we should include this in all our scripts?
We could always add it in the core, but that wouldnt be available until the next full release.

kingthorin:
@shengulong can you test and see if that addresses the issue you were encountering.

If so then please submit a new commit, doing an interactive rebase and fix-up or squash into a single "atomic" commit: https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#what-we-zap-team-expect-from-you .

- https://www.kernel.org/pub/software/scm/git/docs/git-rebase.html#_interactive_mode
- https://wiki.eclipse.org/EGit/User_Guide#Interactive_Rebase

If this is a workable solution I'll add it to my todo list and ensure the other scripts get the same snippet.

thc202:
I'm fine with the change, use of eval function could be dropped though. There are some scripts that might need other tweaks to work as expected (e.g. "Find HTML comments.js", replace print calls with println and remove the newline char).

@shengulong could you also change the commit message to be in English?

kingthorin:
Agreed.

```JavaScript
if (typeof println == 'undefined') this.println = print;

println("Fred");
print("Bob");
```
Is a better solution, avoiding use of evil.....eval :+1:

shengulong:
@kingthorin ，thanks，two ways work well in my environment
@thc202 ok,next time,i will commit message in english
i'm a newer for the zap.
i try to learn the way to use this software for my job

shengulong:
@psiinon I have a question for you,can you answer me?
Now,I want to use zap as my phone's proxy,but failed for the https protocol.The reason is that the zap's caCert that should be installed in my phone.But I cann't find a way to install the ca in my phone like pc.So,can you show the method to install the ca in phone?

kingthorin:
It depends on device and OS specifics. Just hit Google or your favorite search engine.

For Android the following may help, though the specifics may vary due to manyfacturer/provider UI modifications: http://wiki.cacert.org/FAQ/ImportRootCert

kingthorin:
@shengulong are you ok with us closing this PR in favor of implementing the other discussed generic solution?

thc202:
@kingthorin do you plan to also change the community scripts to use the generic solution?

kingthorin:
Yup, that's what I was thinking

thc202:
OK :)

psiinon:
And the templates :)

kingthorin:
Templates are already done :)
https://github.com/zaproxy/zaproxy/pull/3138

psiinon:
\o/

PR #53, opened 2017-02-06
Update README.md

kingthorin:
The persona service has been shutdown. Make a note, and keep the script as an example.

`login.persona.org` reports: 
>The persona.org service has shut down. 

and links to https://wiki.mozilla.org/Identity/Persona_Shutdown_Guidelines_for_Reliers

### DIFF
--- a/standalone/README.md
+++ b/standalone/README.md
@@ -2,3 +2,5 @@ Standalone scripts
 ==================
 
 Scripts that are self contained and are only run when your start them manually. 
+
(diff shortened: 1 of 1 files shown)


standalone/README.md
```diff
@@ -1,4 +1,6 @@
 Standalone scripts
 ==================
 
-Scripts that are self contained and are only run when your start them manually. 
\ No newline at end of file
+Scripts that are self contained and are only run when your start them manually.
```
thc202: Not related to this change but worth fixing(?), there's a typo "your".

kingthorin: Sure, now I just need to figure out how to amend a PR submitted through the web interface :)

kingthorin:
Hmmm well now the PR isn't atomic, can we squash when we merge? That seems to be an option in the web UI now.

thc202:
Done :)

PR #56, opened 2017-04-06
Added showDifferences.js script in httpfuzzerprocessor that compares …

denniskniep:
…the original Response with the fuzzed Response and add the result to the state column.

### DIFF
--- a/httpfuzzerprocessor/showDifferences.js
+++ b/httpfuzzerprocessor/showDifferences.js
@@ -1,0 +1,85 @@
+//A Fuzzer HTTP Processor script that compares the original Response with the fuzzed Response 
+//and add the result to the state column!
+//To remove all other states from the state column set the variable `removeOtherStatesFromStateColumn` to `true`.
+//This might be useful if you want to order the column.
+//Script works only with Java 8 and needs Diff add-on
+ 
+var DiffTool = Java.type("org.zaproxy.zap.extension.diff.diff_match_patch");
+var key = "script.showDifferences.js";
+var showResultInTable = true;
+var removeOtherStatesFromStateColumn = false;
+var original = null;
+
+function processMessage(utils, message) {
+	return message;
+}
+
+// Called after receiving the fuzzed message from the server
+function processResult(utils, fuzzResult){	
+	if(!original){
+		original = responseAsString(utils.getOriginalMessage());
+	}
+	
+	var fuzzed = responseAsString(fuzzResult.getHttpMessage());
+	var diffList = createDiff(original, fuzzed);	
+	var aggregatedDiff = aggregateDiff(diffList);	
+	displayToStateColumn(fuzzResult, aggregatedDiff);
+	return showResultInTable;
+}
+
+function responseAsString(httpMessage){
+	var responseHeader = httpMessage.getResponseHeader().toString();
+	var responseBody = httpMessage.getResponseBody().toString();
+	return responseHeader + "\r\n" + responseBody;
+}
+
+function createDiff(original, fuzzed){
+	var diffTool = new DiffTool();
+	return diffTool.diff_main(original, fuzzed);
+}
+
+function displayToStateColumn(fuzzResult, aggregatedDiff){
+	if(removeOtherStatesFromStateColumn){
+		removeAllStates(fuzzResult);
+	}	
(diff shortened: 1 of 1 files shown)


httpfuzzerprocessor/showDifferences.js
```diff
@@ -0,0 +1,76 @@
+//A Fuzzer HTTP Processor script that compares the original Response with the fuzzed Response 
+//and add the result to the state column!
+ 
+var DiffTool = Java.type("org.zaproxy.zap.extension.diff.diff_match_patch");
+var key = "fuzz.httpfuzzer.messageprocessor.tagcreator.tags.text";
+var showResultInTable = true;
```
thc202: Does it worth having this flag?

httpfuzzerprocessor/showDifferences.js
```diff
@@ -0,0 +1,76 @@
...
+
+
+function processMessage(utils, message) {
+	return message;
+}
+
+// Called after receiving the fuzzed message from the server
+function processResult(utils, fuzzResult){	
+	var original = responseAsString(utils.getOriginalMessage());
```
thc202: This could be done just once.

httpfuzzerprocessor/showDifferences.js
```diff
@@ -0,0 +1,76 @@
...
+	return responseHeader + "\r\n" + responseBody;
+}
+
+function createDiff(original, fuzzed){
+	var diffTool = new DiffTool();
+	return diffTool.diff_main(original, fuzzed);
+}
+
+function displayToStateColumn(fuzzResult, aggregatedDiff){
+	removeAllStates(fuzzResult);
```
thc202: This could be optional (i.e. have a flag to control it), the user might want to have all chosen states in the results.

httpfuzzerprocessor/showDifferences.js
denniskniep: The intention of this named boolean variable is to avoid the comment
` // Whether the result should be accepted, or discarded and not shown.`

But I can remove this variable and return directly `true`...

thc202: Ah, OK, I'm fine using the variable.

httpfuzzerprocessor/showDifferences.js
denniskniep: The Problem is, that `fuzzResult.getCustomStates()` can contain many states, but the  FuzzResultStateHighlighter can not display many states. The Payload Reflection Detector is a default Message Processor and writes always to this column, if he detects something. Than the state of this script will not be displayed.  If the first Message Processor render state in the column no other Message Processor can add something.

I add a flag that can control this behaviour, but the flag is `true` by default. So it remove all states by default.

thc202: Latest version of the fuzzer add-on does show multiple states, they are no longer overwritten (per zaproxy/zaproxy#3166, it just shows one icon though).

httpfuzzerprocessor/showDifferences.js
```diff
@@ -0,0 +1,82 @@
...
+//and add the result to the state column!
+//To remove all other states from the state column set the variable `removeOtherStatesFromStateColumn` to `true`.
+//This might be useful if you want to order the column.
+//Script works only with Java 8
+ 
+var DiffTool = Java.type("org.zaproxy.zap.extension.diff.diff_match_patch");
+var key = "script.showDifferences.js";
+var showResultInTable = true;
+var removeOtherStatesFromStateColumn = false;
+var original = responseAsString(utils.getOriginalMessage());
```
thc202: This should be done in one of the methods, the `utils` is not globally defined.

httpfuzzerprocessor/showDifferences.js
```diff
@@ -0,0 +1,85 @@
+//A Fuzzer HTTP Processor script that compares the original Response with the fuzzed Response 
+//and add the result to the state column!
+//To remove all other states from the state column set the variable `removeOtherStatesFromStateColumn` to `true`.
+//This might be useful if you want to order the column.
+//Script works only with Java 8
```
thc202: Could also mention that if needs Diff add-on :)

thc202:
Thank you!

PR #57, opened 2017-04-17
passive script ->reflected parameter script

deadstar1:
looks for GET parameter values that are reflected in the response.

### DIFF
--- a/passive/find_reflected_params.py
+++ b/passive/find_reflected_params.py
@@ -1,22 +1,19 @@
 """
-
 looks for parameter values that are reflected in the response.
 Author: maradrianbelen.com
-
 The scan function will be called for request/response made via ZAP, excluding some of the automated tools
 Passive scan rules should not make any requests 
-
 Note that new passive scripts will initially be disabled
 Right click the script in the Scripts tree and select "enable"
-
 """  
 def compare(paramvalue_pair,msg): 
-	value=paramvalue_pair.split('=')
-	body=msg.getResponseBody().toString()
-	header=msg.getResponseHeader().toString();
 	reflected=''
-	if (value[1] in body or value[1] in header) and value[1] != '': 
-		reflected=value[0]
+	if len(paramvalue_pair.split('=',1))>1:
+		value=paramvalue_pair.split('=',1)
+		body=msg.getResponseBody().toString()
+		header=msg.getResponseHeader().toString()
+		if (value[1] in body or value[1] in header) and value[1] != '' and len(value[1])>7: 
+			reflected=value
 	return reflected
 
 def scan(ps, msg, src): 
@@ -26,21 +23,12 @@ def scan(ps, msg, src):
 	URI=msg.getRequestHeader().getURI();
 	query=URI.getQuery();
 	if query:
-		print query
 		uriofreflected_param=URI.toString()
 		paramvalue_pair=query.split('&');
 		i=0;
 		while(i<len(paramvalue_pair)):
 			if(compare(paramvalue_pair[i],msg)):
-				reflected_params=reflected_params  + compare(paramvalue_pair[i],msg) + ','
+				reflected_params=reflected_params  + compare(paramvalue_pair[i],msg)[0] + ','
 			i=i+1;
 		if(reflected_params):
 			ps.raiseAlert(0, 2, alertTitle, alertDesc, uriofreflected_param, reflected_params, '', '', '', '', 0, 0, msg);
-
-
-
-
-
-
-
-
--- a/passive/find_reflected_params.py
+++ b/passive/find_reflected_params.py
@@ -1,0 +1,48 @@
+"""
+
+looks for parameter values that are reflected in the response.
+Author: maradrianbelen.com
+
+The scan function will be called for request/response made via ZAP, excluding some of the automated tools
+Passive scan rules should not make any requests 
+
+Note that new passive scripts will initially be disabled
+Right click the script in the Scripts tree and select "enable"
+"""  
+def compare(paramvalue_pair,msg): #get the value in a single parameter and compare it on the HTTP response.
+	value=paramvalue_pair.split('=')
+	#print 'parameter ' + value[0]+ ' is equal to ' + value[1]
+	# add scan for header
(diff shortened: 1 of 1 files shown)


passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+
+The scan function will be called for request/response made via ZAP, excluding some of the automated tools
+Passive scan rules should not make any requests 
+
+Note that new passive scripts will initially be disabled
+Right click the script in the Scripts tree and select "enable"
+"""  
+def compare(paramvalue_pair,msg): #get the value in a single parameter and compare it on the HTTP response.
+	value=paramvalue_pair.split('=')
+	#print 'parameter ' + value[0]+ ' is equal to ' + value[1]
```
thc202: It could be added a flag to print (or not) these debug statements.

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+	# add scan for header
```
thc202: This comment could be removed?

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+	body=msg.getResponseBody().toString()
+	header=msg.getResponseHeader().toString();
+	reflected=''
+	if body.find(value[1])>-1 or header.find(value[1])>-1:
```
thc202: This assumes that the parameter has a value, but it might not, leading to errors.

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+		reflected=value[0]
+	return reflected
+	
+
+
+def scan(ps, msg, src):
+	reflected_params=''
+	URI=msg.getRequestHeader().getURI();
```
thc202: The URI could be used in the other statements (e.g. line 28, 33)

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+	query=msg.getRequestHeader().getURI().getQuery();
+	print 'params of ' + URI.toString();
```
thc202: Worth adding the debug flag for these print statements too.

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+	print '\n'
+	#get a LIST of  param:value pairs. i.e test=ddd&ddd=sdsd
+	if msg.getRequestHeader().getURI().getQuery():
```
thc202: Could use `query` variable?

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+		uriofreflected_param=msg.getRequestHeader().getURI().toString()
+		paramvalue_pair=query.split('&');#test=ddd
+		i=0;
+		while(i<len(paramvalue_pair)): # send a single param:value pair.
+			if(compare(paramvalue_pair[i],msg)):
+				reflected_params=reflected_params + ',' + compare(paramvalue_pair[i],msg)
+			i=i+1;
+		if(reflected_params):
+			ps.raiseAlert(0, 2, 'Find reflected parameter values', 'Reflected parameter value has been found. A reflected parameter values may introduce XSS vulnerability or HTTP header injection.',
```
thc202: The alert name could be "Found reflected..."?

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+			uriofreflected_param,
+			'Reflected Parameters: ' + reflected_params, 'blank', 'blank', '', '', 0, 0, msg);
```
thc202: Better use an empty string instead of adding "blank".

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+	else:
+		print URI.toString() + ' has no parameter'
```
thc202: Same here regarding the debug flag.

passive/find_reflected_params.py
```diff
@@ -0,0 +1,48 @@
...
+
+# line 36-40. the search code does not work properly, and it only shows the last reflected parameter if there is multiple reflected parameters in one HTTP REQUST
+# I am wrong. the reflected_params parameter is incremented when a reflected parameter is found.
```
thc202: Should these comments be removed?

passive/find_reflected_params.py
```diff
@@ -0,0 +1,46 @@
...
+Note that new passive scripts will initially be disabled
+Right click the script in the Scripts tree and select "enable"
+
+"""  
+def compare(paramvalue_pair,msg): 
+	value=paramvalue_pair.split('=')
+	body=msg.getResponseBody().toString()
+	header=msg.getResponseHeader().toString();
+	reflected=''
+	if (value[1] in body or value[1] in header) and value[1] != '':
```
thc202: This will still lead to errors:
```
Traceback (most recent call last):
  File "<script>", line 34, in scan
  File "<script>", line 18, in compare
IndexError: index out of range: 1
```
this can be reproduced with a query parameter without a value, e.g.: `x=y&z`.

passive/find_reflected_params.py
```diff
@@ -0,0 +1,46 @@
...
+	return reflected
+
+def scan(ps, msg, src): 
+	alertTitle='Reflected HTTP GET parameter(s)'
+	alertDesc='Reflected parameter value has been found. A reflected parameter values may introduce XSS vulnerability or HTTP header injection.'
+	reflected_params=''
+	URI=msg.getRequestHeader().getURI();
+	query=URI.getQuery();
+	if query:
+		print query
```
thc202: This could be removed or use a flag?

passive/find_reflected_params.py
```diff
@@ -0,0 +1,46 @@
...
+		if(reflected_params):
+			ps.raiseAlert(0, 2, alertTitle, alertDesc, uriofreflected_param, reflected_params, '', '', '', '', 0, 0, msg);
+
+
+
+
+
+
+
+
```
thc202: Some of the above empty lines could be removed.

thc202:
Thanks!

PR #62, opened 2017-07-12
Extension History Enumerator

acaetano:
The existing search history showed an empty result if the history came from a saved session. 
So this one enumerates all occurrences of a regex match by checking ZAPs DB, displaying it in a text window.

### DIFF
--- a/standalone/extHistoryEnumerator.py
+++ b/standalone/extHistoryEnumerator.py
@@ -1,0 +1,73 @@
+"""
+This script enumerates all occurrences of a text, based on a regex.
+The following search example returns all soapaction names found in 
+    the history, be it in the header or body.
+
+You need to edit the 'regex' variable to whatever suits your need
+"""
+
+from org.parosproxy.paros.model import Model;
+import re;
+from org.parosproxy.paros.view import AbstractFrame;
+from org.zaproxy.zap.utils import ZapTextArea;
+
+sessionId = Model.getSingleton().getSession();
+tbHist = Model.getSingleton().getDb().getTableHistory();
+
+""" Change this variable to match the expression you are looking for """
+regex='soapactio.+';
+
+def search(msg):
+  """ Searching inside the request """
+  match = re.search(regex, msg.getRequestHeader().toString(), re.IGNORECASE);
+  if(match):
+    for header in msg.getRequestHeader().toString().split("\n"):
+      match = re.search(regex, header, re.IGNORECASE);
+      if(match):
+        yield header;
+  match = re.search(regex, msg.getRequestBody().toString(), re.IGNORECASE);
+  if(match):
+    for word in msg.getRequestBody().toString().split():
+      match = re.search(regex, word, re.IGNORECASE);
+      if(match):
+        yield word;
+  """ Searching inside the response """
+  match = re.search(regex, msg.getResponseHeader().toString(), re.IGNORECASE);
+  if(match):
+    for header in msg.getResponseHeader().toString().split("\n"):
+      match = re.search(regex, header, re.IGNORECASE);
+      if(match):
+        yield header;
+  match = re.search(regex, msg.getResponseBody().toString(), re.IGNORECASE);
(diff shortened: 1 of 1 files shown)


standalone/extHistoryEnumerator.py
```diff
@@ -0,0 +1,66 @@
...
+from org.parosproxy.paros.model import Model;
+from org.parosproxy.paros.model import HistoryReference;
+import re;
+from org.parosproxy.paros.view import AbstractFrame;
+from javax.swing import JPanel;
+from org.zaproxy.zap.utils import ZapTextArea;
+
+sessionId = Model.getSingleton().getSession();
+tbHist = Model.getSingleton().getDb().getTableHistory();
+regex='soapaction';
```
kingthorin: Could you add a comment to this line so that it stands out to other users that this is the variable value they need to modify to their own needs. (You could also quote the variable name in the lead-in block comment.) 

Would it be worth making the example be an actual regex? (`'soapactio.*'` or whatever?)

standalone/extHistoryEnumerator.py
```diff
@@ -0,0 +1,66 @@
...
+
+def search(msg):
+  match = re.search(regex, msg.getRequestHeader().toString(), re.IGNORECASE);
```
kingthorin: Might be worth commenting here `Search Request`

standalone/extHistoryEnumerator.py
```diff
@@ -0,0 +1,66 @@
...
+  if(match):
+    for header in msg.getRequestHeader().toString().split("\n"):
+      match = re.search(regex, header, re.IGNORECASE);
+      if(match):
+        yield header;
+  match = re.search(regex, msg.getRequestBody().toString(), re.IGNORECASE);
+  if(match):
+    yield match;
+  match = re.search(regex, msg.getResponseHeader().toString(), re.IGNORECASE);
```
kingthorin: Might be worth commenting here `Search Response`

Just so it sticks out for other users, and doesn't seem like duplicate code.

acaetano:
I found a bug on this PR. Please hold it.

acaetano:
OK, bug fixed and commit squashed. Ready for review.

standalone/extHistoryEnumerator.py
```diff
@@ -0,0 +1,71 @@
+"""
+This script enumerates all occurrences of a text, based on a regex.
+The following search example returns all soapaction names found in 
+    the history, be it in the header or body.
+"""
+
+from org.parosproxy.paros.control import Control;
+from org.parosproxy.paros.extension.history import ExtensionHistory;
+from org.parosproxy.paros.model import Model;
+from org.parosproxy.paros.model import HistoryReference;
```
thc202: There are some unused imports that could be removed?
```Python
from org.parosproxy.paros.control import Control;
from org.parosproxy.paros.extension.history import ExtensionHistory;
from javax.swing import JPanel;
from org.parosproxy.paros.model import HistoryReference;
```

standalone/extHistoryEnumerator.py
```diff
@@ -0,0 +1,71 @@
...
+        yield word;
+  match = re.search(regex, msg.getResponseHeader().toString(), re.IGNORECASE);
+  if(match):
+    for header in msg.getResponseHeader().toString().split("\n"):
+      match = re.search(regex, header, re.IGNORECASE);
+      if(match):
+        yield header;
+  match = re.search(regex, msg.getResponseBody().toString(), re.IGNORECASE);
+  if(match):
+    for word in msg.getRequestBody().toString().split():
```
thc202: `getResponseBody()`

thc202:
Could you provide more details about the search issue you encountered?

acaetano:
Done.

thc202:
Thank you!

kingthorin:
Thanks @acaetano

acaetano:
You're welcome.

PR #65, opened 2017-07-25
add variant script for JSON string input vectors

juhakivekas:
This worked as an input vector when I had to do very careful scanning. The script will recursively tag all JSON stings and later set them using a recursive (but lightweight) function. The use case is kinda specific, but this might be a good sample for anybody making custom JSON input vectors.

### DIFF
--- a/variant/JsonStrings.js
+++ b/variant/JsonStrings.js
@@ -1,0 +1,78 @@
+// The parseParameter function will typically be called for every page and 
+// the setParameter function is called by each active plugin to bundle specific attacks
+
+// Note that new custom input vector scripts will initially be disabled
+// Right click the script in the Scripts tree and select "enable"  
+
+// The following handles differences in printing between Java 7's Rhino JS engine
+// and Java 8's Nashorn JS engine
+if (typeof println == 'undefined') this.println = print;
+
+/*
+This variant script can be used as an alternative to the default JSON input vectors
+when one only wants to scan the string fields of the JSON object, not integers.
+
+Input vectors will be located and set recursively and objects containing up to three
+layers of objects, lists and strings seem to work as intended.
+*/
+
+function parseParameters(helper, msg) {
+	//we're only interested in JSON requests with non-empty POST bodies
+	header = msg.getRequestHeader()
+	if(!header.getHeader("Content-Type")
+		|| !header.getHeader("Content-Type").trim().equalsIgnoreCase("application/json")
+		|| header.getMethod() != "POST"
+		|| msg.getRequestBody().length() == 0
+	){
+		return;
+	}
+
+	body = msg.getRequestBody().toString();
+	try{
+		obj = JSON.parse(body);
+	}
+	catch(err){
+		print("Parsing message body failed: " + err.message);
+		return;
+	}
+	recursive_parse(helper, obj, "Object");
+}
+
+function recursive_parse(helper, obj, path){
+	if (typeof(obj) == "number"){
+		//skipping this is why I wrote this script!
+	}else if(typeof(obj) == "string"){
+		//here the parameter is added with the name being a JSON
+		//list of keys that act as the JSON path to the input vector
+		helper.addParamPost(path, obj);
+	}else {
+		for (k in obj){
+			recursive_parse(helper, obj[k], path + "." + k)
+		}
+	}
+	//are there other cases that have been missed?
+}
+function setParameter(helper, msg, param, value, escaped) {
+	try{
+		obj = JSON.parse(msg.getRequestBody().toString());
+		//get the path from the parameter name, ignore the "Object"-part
+		path = param.split('.').slice(1);
+		//this may fail if "param" has been edited since parsing
+		recursive_set(obj, path, value);
+	}
+	catch (err){
+		print("Setting parameter value failed: " + err.message);
+		return;
+	}
+	msg.getRequestBody().setBody(JSON.stringify(obj));
+}
+
+function recursive_set(obj, path, value){
+	//print(JSON.stringify(obj) + " : " + JSON.stringify(path))
+	if (path.length == 0){
+		//stop recursing when the path is consumed
+		return value;
+	}
+	obj[path[0]] = recursive_set(obj[path[0]], path.slice(1), value);
+	return obj;
+}


variant/JsonStrings.js
```diff
@@ -0,0 +1,61 @@
...
+This variant script can be used as an alternative to the default JSON input vectors
+when one only wants to scan the string fields of the JSON object, not integers.
+
+Input vectors will be located and set recursively and objects containing up to three
+layers of objects, lists and strings seem to work as intended.
+*/
+
+function parseParameters(helper, msg) {
+	//we're only interested in JSON requests
+	if (msg.getRequestHeader().getHeader("Content-Type") != 'application/json'){
```
kingthorin: Might be "safest" to `trim()` the header value, then use `equalsIgnoreCase()` just on the chance that the script hits something like:

- Application/json
- application/JSON

variant/JsonStrings.js
```diff
@@ -0,0 +1,61 @@
...
+		return;
+	}
+
+	body = msg.getRequestBody().toString();
+	if (body == "") {return;}
+	obj = JSON.parse(body)
```
thc202: Worth adding a try/catch to avoid "breaking" the script if the content is malformed.

variant/JsonStrings.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function recursive_parse(helper, obj, path){
+	if (typeof(obj) == "number"){
+		//skipping this is why I wrote this script!
+	}else if(typeof(obj) == "string"){
+		//print(JSON.stringify(path) + " : " + obj)
+		//here the parameter is added with the name being a JSON
+		//list of keys that act as the JSON path to the input vector
+		helper.addParamPost(JSON.stringify(path), obj);
```
thc202: Ideally the name should be just the name of the parameter/field/property, but with current script input vector implementation it's not possible to correctly identify the field being changed...

juhakivekas: Would `Object.first.second.third` be ok for this?

thc202: Unfortunately there's no workaround for this issue, leaving as is is fine (we would have to change the script interface to fix this). Worth adding a try/catch here as well though, just in case a scanner is messing with the name.

juhakivekas:
screen shot 2017-07-25 at 16 50 47
 That does look pretty ok to me though. Is there a reason this should be more readable tahn the fact that it should be more informative?

Also what exactly should the `catch` do? If i print the error message and return I get a really silent failure. Is that the point?

variant/JsonStrings.js
juhakivekas: screen shot 2017-07-25 at 16 50 47

 That does look pretty ok to me though. Is there a reason this should be more readable than the fact that it should be more informative?

Also what exactly should the `catch` do? If i print the error message and return I get a really silent failure. Is that the point?

thc202: That's another case, I was not referring to the UI/Alert data but the behaviour of the input vector script, both name and value can be changed by the scanners, so, if a scanner sets the name to other thing it will break this input vector script, but again there's no workaround to properly identify the parameter being changed.

Yes, printing something (or log) and return is the best that the script can do IMO, the alternative is to just let ZAP disable the script because of the error.

kingthorin:
Please do an interactive rebase and fix-up or squash into a single "atomic" commit: https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#what-we-zap-team-expect-from-you .

- https://www.instructables.com/id/How-to-Rebase-Interactive-With-Eclipse-eGit/
- https://www.kernel.org/pub/software/scm/git/docs/git-rebase.html#_interactive_mode
- https://wiki.eclipse.org/EGit/User_Guide#Interactive_Rebase

variant/JsonStrings.js
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	body = msg.getRequestBody().toString();
+	try{
+		obj = JSON.parse(body);
+	}
+	catch(err){
+		print("Parsing message body failed: " + err.message)
+	}
+	recursive_parse(helper, obj, "Object");
```
thc202: This should be moved to inside the try block (`obj` is not be defined in case of a parsing error).

variant/JsonStrings.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function recursive_parse(helper, obj, path){
+	if (typeof(obj) == "number"){
+		//skipping this is why I wrote this script!
+	}else if(typeof(obj) == "string"){
+		//here the parameter is added with the name being a JSON
+		//list of keys that act as the JSON path to the input vector
+		helper.addParamPost(JSON.stringify(path), obj);
```
thc202: `JSON.stringify` no longer need here, right? Otherwise it adds quotes to the path (e.g. `"Object.a"` instead of just `Object.a`) leading to wrong path when setting the value.

variant/JsonStrings.js
```diff
@@ -0,0 +1,71 @@
...
+		obj = JSON.parse(msg.getRequestBody().toString());
+		//get the path from the parameter name, ignore the "Object"-part
+		path = param.split('.').slice(1);
+		//this may fail if "param" has been edited since parsing
+		recursive_set(obj, path, value);
+	}
+	catch (err){
+		print("Setting parameter value failed: " + err.message)
+	}
+	msg.getRequestBody().setBody(JSON.stringify(obj));
```
thc202: Same here (re the try block).

variant/JsonStrings.js
juhakivekas: I added a return in the try instead

variant/JsonStrings.js
juhakivekas: Noticed this too during work, fixed now

juhakivekas:
Is that the right kind of rebased commit?

variant/JsonStrings.js
juhakivekas: added return here again

kingthorin:
Looks good to me.

Yes if you're going to contribute something it is generally best to create a "feature branch" (based on current master/develop) and do your PR from that. (There's a summary here: https://github.com/zaproxy/zaproxy/wiki/Contributing-Changes)

variant/JsonStrings.js
```diff
@@ -0,0 +1,80 @@
...
+*/
+
+function parseParameters(helper, msg) {
+	//we're only interested in non-empty JSON requests
+	header = msg.getRequestHeader()
+	if(
+		    header.getHeader("Content-Type")
+		&& !header.getHeader("Content-Type").trim().equalsIgnoreCase("application/json")
+		&&  header.getMethod() == "POST"
+		&&  msg.getRequestBody().length() > 0
```
thc202: The whole condition does not look right. Should use ors and negate some of the conditions?

juhakivekas: This pull request is starting to give away my non-developerness so hard 😂

kingthorin: MEH, you should read through some of mine..... nothing to worry about. Just helps us up our game 🥇

variant/JsonStrings.js
```diff
@@ -0,0 +1,77 @@
...
+when one only wants to scan the string fields of the JSON object, not integers.
+
+Input vectors will be located and set recursively and objects containing up to three
+layers of objects, lists and strings seem to work as intended.
+*/
+
+function parseParameters(helper, msg) {
+	//we're only interested in non-empty JSON requests
+	header = msg.getRequestHeader()
+	if(header.getHeader("Content-Type") && !header.getHeader("Content-Type").trim().equalsIgnoreCase("application/json")
```
thc202: First expression should be negated and should be used or instead of and?

thc202: @juhakivekas would you be available to take a look at this? Otherwise we can tweak it if not.

thc202:
Thanks!

PR #67, opened 2017-08-15
Payload generator template for request fields that are somewhat related.

acaetano:
Also an example on solving some strange Jython issues regarding type conversion.

### DIFF
--- a/payloadgenerator/associated_fields.py
+++ b/payloadgenerator/associated_fields.py
@@ -1,0 +1,41 @@
+"""
+This script is just a proof of concept for payload generation, 
+supplementing ZAP's native generation mechanism(s).
+It generates payload1 for field1 and calculates the payload 
+(payload2 = sha1(base64(payload1))) for field2, something like Battering Ram
+without the hassle of generating the payload2 wordlist first
+"""
+
+from base64 import b64encode;
+from hashlib import sha1;
+from java.lang import Long;     # bug resolution import
+
+INITIAL_VALUE = 0; 
+count = INITIAL_VALUE;
+passfile_path = 'K:\\payloads.txt';
+pl_buff = list();
+lines = 0;
+for line in open(passfile_path):  
+    pl_buff.append(line.rstrip());
+    lines += 1;
+middle = str('&check=');
+NUMBER_OF_PAYLOADS = lines;
+
+def getNumberOfPayloads():
+    return Long(NUMBER_OF_PAYLOADS);    # Typecast to solve bug
+
+def hasNext():
+    return (count < NUMBER_OF_PAYLOADS);
+
+def next():
+    global count;
+    num = count;
+    count+=1;
+    return pl_buff[num]+middle+sha1(b64encode(pl_buff[num])).hexdigest();
+
+def reset():
+    count = INITIAL_VALUE;
+
+def close():
+    pass;
+


payloadgenerator/associated_fields.py
```diff
@@ -0,0 +1,33 @@
+from base64 import b64encode;
```
thc202: It would be great if it had a description describing what the generator does to the payloads.

payloadgenerator/associated_fields.py
```diff
@@ -0,0 +1,33 @@
...
+pl_buff = list();
+lines = 0;
+for line in open(passfile_path):  
+    pl_buff.append(line.rstrip());
+    lines += 1;
+middle = str('&check=');
+NUMBER_OF_PAYLOADS = lines;
+
+def getNumberOfPayloads():
+    return Long(NUMBER_OF_PAYLOADS);    # Typecast to solve bug
```
thc202: Is there a Jython issue for this?

payloadgenerator/associated_fields.py
acaetano: It just generates payload1 for a field1 and calculates the payload (payload2 = sha1(payload1)) for field2. Actually, it is just a concept about how to workaround the problem of not having other types of attack like burp, as ZAP natively only implements Sniper mode. 

This script tries to implement the BatteringRam idea, but it actually came out better than that, as the second payload can produce something that is related to payload1 but still different, without the need of creating a previous wordlist of sha1(payload1) like Burp does.

acaetano:
Description added.

payloadgenerator/associated_fields.py
```diff
@@ -0,0 +1,42 @@
+"""
+This script is just a concept about how to workaround the problem 
+of not having other types of attack like burp, as ZAP natively 
+only implements Sniper mode.
```
kingthorin: Instead of promoting product comparison, how about something like:
```
This script is just a proof of concept for payload generation, 
supplementing ZAP's native generation mechanism(s).
```

payloadgenerator/associated_fields.py
```diff
@@ -0,0 +1,41 @@
+"""
+This script is just a proof of concept for payload generation, 
+supplementing ZAP's native generation mechanism(s).
+It generates payload1 for field1 and calculates the payload 
+(payload2 = sha1(payload1)) for field2, something like Battering Ram
```
thc202: Should mention that the `payload1` is Base64 encoded before being hashed?

payloadgenerator/associated_fields.py
thc202: Thanks.

payloadgenerator/associated_fields.py
thc202: Right, I was asking if there was an issue in the Jython project about this, maybe http://bugs.jython.org/issue1494 ? that seems to be fixed in newer versions (following other issue reports). Did you try with the new Jython version?

acaetano:
No, the bug persists even after Jython update.

thc202:
OK.

Do you plan to tweak the description of the script (https://github.com/zaproxy/community-scripts/pull/67#discussion_r133665657)?

acaetano:
Oops. Description updated.

thc202:
Thank you!

PR #73, opened 2017-10-27
Added extender directory

psiinon:


kingthorin:
Here too?
https://github.com/zaproxy/zaproxy/tree/develop/src/scripts/templates

psiinon:
I was actually thinking we should move these from the core into the scripts add-on - that way we can patch them without re-releasing the core.
wdyt?
In any case, the core doesnt actually know about extender scripts, they are just defined in the scripts addon.

kingthorin:
Moving them sounds good to me.

thc202:
I'm fine with the move as well.

thc202:
We could also move the script extension into an add-on, not just its templates.

PR #76, opened 2017-11-09
Added a script to collect all unique past cookies and show them in a window.

acaetano:


acaetano:
I tried to group them by site, but the HttpCookie object had no domain information. The issue with a targeted version is that i could only scan a single URL. 

Right? Or is it possible to scan a domain using the Sites tree?

standalone/past_cookies_jar.py
```diff
@@ -0,0 +1,47 @@
...
+from org.parosproxy.paros.view import AbstractFrame;
+from org.zaproxy.zap.utils import ZapTextArea;
+from javax.swing import JScrollPane;
+
+sessionId = Model.getSingleton().getSession();
+tbHist = Model.getSingleton().getDb().getTableHistory();
+
+def collect(msg):
+  """ Collecting cookie from HttpMessage """
+  cookie = msg.getRequestHeader().getHttpCookies();
```
thc202: This could be renamed to `cookies`.

standalone/past_cookies_jar.py
```diff
@@ -0,0 +1,47 @@
...
+    self.setSize(700, 500);
+    ta = ZapTextArea(text);
+    sp = JScrollPane(ta, 
+		JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, 
+		JScrollPane.HORIZONTAL_SCROLLBAR_AS_NEEDED);
+    self.add(sp);
+    self.setVisible(True);
+    
+if (tbHist != None):
+  final = list();
```
thc202: `uniqueCookies`?

thc202:
It can be extracted from the message itself (`msg.getRequestHeader().getURI().getHost()`?).

acaetano:
Changes made. You're right about taking it from the url. I added that as well.

The only issue is that if the site defined the cookie with a different scope (like site is www. but cookie url is api.), we won't know it. But this is better than not knowing anything at all.

Ready for new review.

standalone/past_cookies_jar.py
```diff
@@ -0,0 +1,60 @@
...
+  for index in tbHist.getHistoryIds(sessionId.getSessionId()):
+    try:
+      msg = tbHist.read(index).getHttpMessage();
+      results = collect(msg);
+      if results is None:
+        continue;
+      for item in results:
+        item.setDomain(msg.getRequestHeader().getURI().getHost());
+        match = re.search(cookie_domain_regex, item.getDomain(), re.IGNORECASE);
+        if(match):
```
thc202: This check could be done outside/before the for loop.

thc202:
Right, you would need to check the cookies when set (e.g. `HttpResponseHeader.getHttpCookies(String)`), or, access the global `HttpState` (but that might not be always enabled).

standalone/past_cookies_jar.py
acaetano: Ok, i could get the domain out of the loop, but i need to set and match it inside the loop.

thc202: Does the match also need to be? The host is the same for all the cookies being checked in the loop, so it could be done just once before. Or is that to be easier to change if/when the cookies have other domain than the message?

kingthorin:
@acaetano are you able to provide updates or answers to @thc202’s comments?

acaetano:
A few optimizations. It is now getting the domain out of the loop and referencing it through a variable, instead of making various "item.GetDomain()" calls.

Ready for review. Sorry for the delay, it got lost somewhere in my stack of things to do. I think i will have to make a notebook for that...

thc202:
Thanks!

PR #79, opened 2018-01-09
Update MediaWiki Auth Script Per User Group

kingthorin:
Per
https://groups.google.com/forum/#!topic/zaproxy-users/ijDxvsZI6do
ensure Content-Length is set appropriately.

### DIFF
--- a/authentication/MediaWikiAuthentication.js
+++ b/authentication/MediaWikiAuthentication.js
@@ -94,6 +94,7 @@ MWAuthenticator.prototype = {
 			requestInfo += ' with body: ' + requestBody;
 			msg.setRequestBody(requestBody);
 		}
+		msg.getRequestHeader().setContentLength(msg.getRequestBody().length());
 
 		println(requestInfo);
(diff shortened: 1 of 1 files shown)


authentication/MediaWikiAuthentication.js
```diff
@@ -94,6 +94,7 @@ MWAuthenticator.prototype = {
 			requestInfo += ' with body: ' + requestBody;
 			msg.setRequestBody(requestBody);
 		}
+		msg.getRequestHeader().setContentLength(requestBody.length);
```
thc202: Better with `msg.getRequestBody().length()` (avoids byte vs char issues and good example for other scripts).

kingthorin: Ok I'll tweak that.

thc202:
Do you plan to update the other scripts?

kingthorin:
I was kind of after a quick hit. This was the one mentioned so I tackled it.

thc202:
OK, we can tackle the others later.

PR #83, opened 2018-01-22
Tweak passive scanner F5 BigIP

thc202:
Change f5_bigip_cookie_internal_ip.js to check all bigip cookies, even
if one of them does not have the expected contents. Also, when handling
IPv6 address do not compress its zeros, it's not working properly (the
resulting address could have more than one "::", which is not valid).


thc202:
No, those happened while testing the migration to Nashorn. I used the IPv6 address in the comment to test it and that lead to the issues (invalid IPv6 and no cookies further processed).

kingthorin:
Oops...bad connectivity.

kingthorin:
Ok, thanks for the tweak.

PR #87, opened 2018-02-02
Correct lower case comparisons

thc202:
Change usage of String.toLowerCase() to use a neutral Locale to avoid
issues with languages that have different characters for lower case.

Related to zaproxy/zaproxy#4327 - ZAP relies on default locale when it
shouldn't


psiinon:
Looks good but now has conflicts

thc202:
Done.

PR #88, opened 2018-02-02
Update JavaScript scripts to Nashorn

thc202:
Update JavaScript scripts (and examples in READMEs) to Nashorn:
 - Remove Nashorn `println` workarounds;
 - Remove comments that mentioned Java 8 or Nashorn as requirement, now
   all of them require the newer engine (per ZAP targeted Java version);
 - Use `print` instead of `println`;
 - Import classes using `Java.type`;
 - Correct declaration of variables in the same statement, fix loops and
   char comparisons;
 - Correct access to constants/methods from extending classes (e.g.
   `HttpHeader` instead of `HttpRequestHeader`, `InetAddress` instead of
   `Inet4Address`);
 - Correct access to classes (e.g. `Model` instead of `Model()`) and
   instantiations (e.g. use `new`).

Related to zaproxy/zaproxy#3470 - Move to using Java 8.

### DIFF
--- a/authentication/MediaWikiApiAuthentication.js
+++ b/authentication/MediaWikiApiAuthentication.js
@@ -23,15 +23,12 @@
  * @author grunny
  */
  
-// The following handles differences in printing between Java 7's Rhino JS engine
-// and Java 8's Nashorn JS engine
-if (typeof println == 'undefined') this.println = print;
+var HttpRequestHeader = Java.type("org.parosproxy.paros.network.HttpRequestHeader");
+var HttpHeader = Java.type("org.parosproxy.paros.network.HttpHeader");
+var URI = Java.type("org.apache.commons.httpclient.URI");
 
 function authenticate(helper, paramsValues, credentials) {
-	println("Authenticating via JavaScript script...");
-	importClass(org.parosproxy.paros.network.HttpRequestHeader);
-	importClass(org.parosproxy.paros.network.HttpHeader);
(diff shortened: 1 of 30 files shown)


authentication/MediaWikiApiAuthentication.js
```diff
@@ -87,16 +84,16 @@ MWApiAuthenticator.prototype = {
 
 	doRequest: function (url, requestMethod, requestBody) {
 		var msg,
-			requestUri = new URI(url, false);
+			requestUri = new URI(url, false),
```
kingthorin: Comma?

authentication/MediaWikiAuthentication.js
```diff
@@ -83,7 +80,7 @@ MWAuthenticator.prototype = {
 	doRequest: function (url, requestMethod, requestBody) {
 		var msg,
 			requestInfo,
-			requestUri = new URI(url, false);
+			requestUri = new URI(url, false),
```
kingthorin: Comma?

authentication/MediaWikiApiAuthentication.js
thc202: Yeah, this is only one statement declaring 3 variables.

authentication/MediaWikiAuthentication.js
thc202: Ditto (with 4 vars).

PR #89, opened 2018-02-10
Added small usage comment

alexlauerman:
Feel free to reject if this is too small.


alexlauerman:
Yes, the file is called Return fake response.js. The response can be overridden in either function.

If you put the response overwrite logic in the proxyRequest(), it does not send the request to the server at all. If you put the logic in proxyResponse(), it sends the request to the server, and overrides the response. 

In my case I was changing the result from a POST, but the POST was never sent if the logic is in proxyRequest, which is not how I thought it would work, so I added a comment in case others run into this. If it's still unclear, feel free to reject. It's a pretty small change.

thc202:
Thanks!

PR #98, opened 2018-03-02
Create loadListInGlobalVariable.js

Naramsim:
Add a simple example to load a list in a global variable

### DIFF
--- a/standalone/loadListInGlobalVariable.js
+++ b/standalone/loadListInGlobalVariable.js
@@ -1,13 +1,8 @@
 //@zaproxy-standalone
 
-/*
-Note that the script below will work only on Zaproxy > 2.7.0
-Since .getGlobalVars() is not supported on previous versions
-*/
-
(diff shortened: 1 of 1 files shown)


standalone/loadListInGlobalVariable.js
```diff
@@ -0,0 +1,9 @@
+//@zaproxy-standalone
+
+org.zaproxy.zap.extension.script.ScriptVars.setGlobalVar('LIST', JSON.stringify(['Zaproxy', 'Zap', 'Simon', 'Mozilla']))
+
+list = JSON.parse(org.zaproxy.zap.extension.script.ScriptVars.getGlobalVars()['LIST'])
```
thc202: Might worth adding a comment saying that this will work only with ZAP < 2.7.0 (or use `getGlobalVar('LIST')`?).

Naramsim: Yes sure, I didn't know about it

thc202:
Thank you!

PR #99, opened 2018-03-02
Create dropCookiesSelectively.js

Naramsim:
Add a simple example to drop from Requests selected cookies

### DIFF
--- a/proxy/dropCookiesSelectively.js
+++ b/proxy/dropCookiesSelectively.js
@@ -3,18 +3,16 @@
 globalCookies = ['CONSENT', 'GZ', 'SID']
 
 function proxyRequest(msg) {
-    globalCookies.forEach(function (cookieToDrop) {
-        cookies = msg.getRequestHeader().getHttpCookies() // This is a List<HttpCookie>
-        iterator = cookies.iterator()
-        while (iterator.hasNext()) {
-            cookie = iterator.next() // This is a HttpCookie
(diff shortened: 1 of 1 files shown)


proxy/dropCookiesSelectively.js
```diff
@@ -0,0 +1,23 @@
+//@zaproxy-proxy
+
+globalCookies = ['CONSENT', 'GZ', 'SID']
+
+function proxyRequest(msg) {
+    globalCookies.forEach(function (cookieToDrop) {
+        cookies = msg.getRequestHeader().getHttpCookies() // This is a List<HttpCookie>
```
thc202: This could be called outside the function, to avoid reprocess the cookies each time.

proxy/dropCookiesSelectively.js
```diff
@@ -0,0 +1,23 @@
...
+        iterator = cookies.iterator()
+        while (iterator.hasNext()) {
+            cookie = iterator.next() // This is a HttpCookie
+            if (cookie.name == cookieToDrop) {
+                iterator.remove()
+                print('Stripped away: ' + cookie.name)
+            }
+        }
+        msg.getRequestHeader().setCookies(cookies)
```
thc202: Per above comment, this could be called outside the function and only if there was any change (no need to re-set them if the same).

thc202:
Thanks!

PR #102, opened 2018-05-08
Add Alerts & Plugin Details Standalone Script

kingthorin:
Per https://stackoverflow.com/questions/50187141/how-can-i-can-list-of-alerts-associated-with-scan-rules-in-owasp-zap

### DIFF
--- a/standalone/alertAndPluginDetails.js
+++ b/standalone/alertAndPluginDetails.js
@@ -1,0 +1,53 @@
+/* Output sample:
+ *     Alert_Name	Source:PluginName	WASC	CWE
+ *     Cross Site Scripting (DOM Based)	ACTIVE:Cross Site Scripting (DOM Based)	8	79
+ *     Non-Storable Content	PASSIVE:Content Cacheability	13	524
+ * It's tab separated so you can simply copy/paste it into Excel (or whatever).
+ */
+
+extAlert = org.parosproxy.paros.control.Control.getSingleton().
+    getExtensionLoader().getExtension(
+        org.zaproxy.zap.extension.alert.ExtensionAlert.NAME) 
+
+extPscan = org.parosproxy.paros.control.Control.getSingleton().
+    getExtensionLoader().getExtension(
+        org.zaproxy.zap.extension.pscan.ExtensionPassiveScan.NAME);
+
+var pf = Java.type("org.parosproxy.paros.core.scanner.PluginFactory");
+
+printHeaders();
+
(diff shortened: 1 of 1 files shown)


standalone/alertAndPluginDetails.js
```diff
@@ -0,0 +1,40 @@
+extAlert = org.parosproxy.paros.control.Control.getSingleton().
+    getExtensionLoader().getExtension(
+        org.zaproxy.zap.extension.alert.ExtensionAlert.NAME) 
+
+extAscan = org.parosproxy.paros.control.Control.getSingleton().
```
thc202: Remove? Not used.

standalone/alertAndPluginDetails.js
```diff
@@ -0,0 +1,40 @@
...
+
+extPscan = org.parosproxy.paros.control.Control.getSingleton().
+    getExtensionLoader().getExtension(
+        org.zaproxy.zap.extension.pscan.ExtensionPassiveScan.NAME);
+
+var pf = Java.type("org.parosproxy.paros.core.scanner.PluginFactory");
+
+printHeaders();
+
+if (extAlert != null) {
```
thc202: ` && extPscan != null`? or later in the passive check?

standalone/alertAndPluginDetails.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+function printHeaders() {
+    print('AlertName\tSource:PluginName\tWASC\tCWE');
+}
+
+function printAlert(alert) {
+    var scanner = '';
+    if (alert.getSource() == Alert.Source.ACTIVE) {
+        scanner = pf.getLoadedPlugin(alert.getPluginId()).getName();
```
thc202: It should check that the scanner exists before getting the name (same for passive), it might happen that the corresponding add-on is not installed.
```
TypeError: null has no such function "getName"
<eval>:33 TypeError: null has no such function "getName"
```

thc202:
It might worth adding a description of the what the output/format is.

standalone/alertAndPluginDetails.js
kingthorin: Added later as suggested.

kingthorin:
Done

thc202:
Thank you!

kingthorin:
Added an ACTIVE entry to the sample output at the top.

thc202:
Still looking good :)

PR #104, opened 2018-05-23
Export Websocket Communication to Emacs Org File

ManosMagnus:
This script export WebSocket communication to [Emacs Org](https://orgmode.org/) File .
Script require the external Module [PyOrgMode](https://github.com/bjonnh/PyOrgMode) 

Export Format:
```
* (#<ChannelID>) <WebSocket Channel host name>
** Switching Protocol
*** Handshake Request
	#+BEGIN_SRC html
	<Http Handshake Request>
	#+END_SRC
*** Handshake Respone
	#+BEGIN_SRC html
	<Http Handshake Request>
	#+END_SRC
** Messages
*** (#<ChannelID>.<MessageID) <Direction> <Opcode> <Date Time> <Byte Lenght>
	- Text
	<Text Payload>
	- Hex
	<Hex Payload>
```
If we will have problem to use external modules, I could implement my own functions in script. If you find this script unnecessary please close the PR

### DIFF
--- a/standalone/WebSocketExportToOrg.py
+++ b/standalone/WebSocketExportToOrg.py
@@ -1,0 +1,141 @@
+"""
+This script export WebSocket communication to Emacs Org File.
+Script require the external Module PyOrgMode (https://github.com/bjonnh/PyOrgMode)
+To add an external module: Tools -> Options -> Jython 
+
+Export Format:
+
+* (#<ChannelID>)<WebSocket Channel host name>
+** Summary [Optional]
+	 - Host: <host>
+	 - URI: <URI>
+	 - Port: <port>
+	 - Websocket Port: <WebSocket Port>
+	 - Websocket URL: <WebSocket URL>
+	 - Messages Opcodes: 
+	 - Outgoing Messages format: 
+	 - Incoming Messages format: 
+	 - Input Messages encryption: 
+	 - Output Messages encryption: 
+	 - Ping/Pong: 
+	 - Close message: Opcode= , Text= 
+** Switching Protocol
+*** Handshake Request
+	#+BEGIN_SRC html
+	<Http Handshake Request>
+	#+END_SRC
+*** Handshake Respone
+	#+BEGIN_SRC html
+	<Http Handshake Request>
+	#+END_SRC
+** Messages
+*** (#<ChannelID>.<MessageID) <Direction> <Opcode> <Date Time> <Byte Lenght>
+	- Text
+	<Text Payload>
+	- Hex
+	<Hex Payload>
+"""
+
+from org.parosproxy.paros.control import Control
+from org.parosproxy.paros.extension.history import ExtensionHistory
+from org.zaproxy.zap.extension.websocket import ExtensionWebSocket
+from org.zaproxy.zap.extension.websocket import WebSocketChannelDTO
+from org.zaproxy.zap.extension.websocket import WebSocketMessageDTO
+from org.zaproxy.zap.extension.websocket.utility import InvalidUtf8Exception
+
+import PyOrgMode
+import sys
+import binascii
+
+reload(sys)
+sys.setdefaultencoding('utf-8')
+
+extWebSocket = Control.getSingleton().getExtensionLoader().getExtension(ExtensionWebSocket.NAME)
+pathToOrgFile = "/path/to/file.org"
+printSummary = True
+base = PyOrgMode.OrgDataStructure()
+channelsList = extWebSocket.getStorage().getTable().getChannelItems()
+
+# Loop through the Channel List
+for channel in channelsList:
+  hr = channel.getHandshakeReference()
+  newChannelElement = PyOrgMode.OrgNode.Element()
+  newChannelElement.level = 1
+  newChannelElement.heading = "(#"+str(channel.id)+"): "
+  if(hr != None):
+    handshakeMessage = hr.getHttpMessage()    
+    newChannelElement.heading =  newChannelElement.heading + handshakeMessage.getRequestHeader().getHostName() + "\n"
+    #Print Summary
+    if(printSummary):
+      newSummary = PyOrgMode.OrgNode.Element() 
+      newSummary.heading = "Summary"
+      newSummary.level = 2
+      newSummary.append("\t - Host: " + handshakeMessage.getRequestHeader().getHostName() + "\n")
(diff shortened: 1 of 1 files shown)


standalone/WebSocketExportToOrg.py
```diff
@@ -0,0 +1,109 @@
...
+extWebSocket = Control.getSingleton().getExtensionLoader().getExtension(ExtensionWebSocket.NAME)
+extHist = Control.getSingleton().getExtensionLoader().getExtension(ExtensionHistory.NAME)
+
+base = PyOrgMode.OrgDataStructure()
+channelId = 0
+if(extHist != None):
+  i=1
+  # Loop through the history table, printing out the history id and the URL
+  hr = extHist.getHistoryReference(i)
+  while(hr != None):
```
psiinon: We can have null HistoryRecords without them being at the end.
https://github.com/zaproxy/community-scripts/blob/master/standalone/Loop%20through%20history%20table.js#L11-L13 is the best way to check for the last record

standalone/WebSocketExportToOrg.py
```diff
@@ -0,0 +1,109 @@
...
+        newMessage.append("\t - HEX\n \t #+BEGIN_SRC\n " + binascii.hexlify(messageUTF8) + "\n\t #+END_SRC\n\n")
+        newMessages.append_clean(newMessage)
+      
+      newChannelElement.append_clean(newMessages)
+      base.root.append_clean(newChannelElement)
+      
+    i += 1
+    hr = extHist.getHistoryReference(i)
+
+base.save_to_file("/path/to/file.org")
```
thc202: This could use a variable defined at the top (easier to notice/change).

standalone/WebSocketExportToOrg.py
thc202: It could also obtain the channels directly (e.g. `extWebSocket.getStorage().getTable().getChannelItems()`) instead of recreating them based on the HTTP message (the handshake can be obtained from the channel, if it still exists).

ManosMagnus: @thc202 That will work also with closed connections?

ManosMagnus:
Suggestions addressed. Also add an optional summary section

standalone/WebSocketExportToOrg.py
```diff
@@ -0,0 +1,141 @@
...
+
+import PyOrgMode
+import sys
+import binascii
+
+reload(sys)
+sys.setdefaultencoding('utf-8')
+
+extWebSocket = Control.getSingleton().getExtensionLoader().getExtension(ExtensionWebSocket.NAME)
+extHist = Control.getSingleton().getExtensionLoader().getExtension(ExtensionHistory.NAME)
```
thc202: Could be removed, no longer used (same for the corresponding import).

standalone/WebSocketExportToOrg.py
```diff
@@ -0,0 +1,141 @@
...
+pathToOrgFile = "/path/to/file.org"
+printSummary = True
+base = PyOrgMode.OrgDataStructure()
+channelsList = extWebSocket.getStorage().getTable().getChannelItems()
+
+for channel in channelsList:
+  handshakeMessage = channel.getHandshakeReference().getHttpMessage()
```
thc202: Note that the `handshakeMessage` will be null if the message was deleted (not usual but can happen).

standalone/WebSocketExportToOrg.py
```diff
@@ -0,0 +1,141 @@
...
+    newSummary.append("\t - Messages Opcodes: \n")
+    newSummary.append("\t - Outgoing Messages format: \n")
+    newSummary.append("\t - Incoming Messages format: \n")
+    newSummary.append("\t - Input Messages encryption: \n")
+    newSummary.append("\t - Output Messages encryption: \n")
+    newSummary.append("\t - Ping/Pong: \n")
+    newSummary.append("\t - Close message: Opcode= , Text= \n")
+    newChannelElement.append_clean(newSummary);
+
+  print(handshakeMessage.getRequestHeader().getHostName())
```
thc202: Remove?

ManosMagnus: I'am not sure about. I think it is better now, make clear that something happened

thc202: Makes sense, my initial impression was that this was a debug statement, worth adding "Processing [something]" (better to use the channel ID? it's guaranteed to exist) to be more obvious that's reporting progress, also worth adding a message at end to let know it finished.

standalone/WebSocketExportToOrg.py
```diff
@@ -0,0 +1,142 @@
...
+    newHandShake.append_clean(newHandShakeResponse)
+
+    newChannelElement.append_clean(newHandShake)
+    
+  print("Processing Channel ID: #" + str(channel.id))
+  
+  #Get Messages  
+  messageCriteria = WebSocketMessageDTO(channel)
+
+  #messagesList = extWebSocket.getWebsocketMessages(messageCriteria,None, None,None,0,0,10000) #Take a look at https://github.com/zaproxy/zap-extensions/pull/15065
```
thc202: There's a typo in the pull request number, but better restore the old method instead, to avoid breaking the script.

standalone/WebSocketExportToOrg.py
```diff
@@ -0,0 +1,142 @@
...
+    newMessage = PyOrgMode.OrgNode.Element()
+    newMessage.level = 3
+    newMessage.heading = "(#" + str(channel.id) + "." + str(message.id) +") "+ direction + " " + message.readableOpcode +" "+ message.dateTime +" "+ str(message.payloadLength) 
+
+    #messageUTF8 = message.getPayloadAsString() #Take a look at https://github.com/zaproxy/zap-extensions/pull/1637
+    try:
+      messageUTF8 = messages.getReadablePayload()
+    except:
+      messageUTF8 = "<unreadable binary payload>"
+    messageHex = message.payload
```
thc202: Just noticed that this is not being used, but should when writing the message in hex?

ManosMagnus: Do you mean that will be better to use messageHex instead of binascii.hexlify(messageUTF8) ?

thc202: It's preferable to use `messageHex` instead of `messageUTF8` (the latter might not have the actual contents), independently how it's saved later.

thc202:
Thanks!

ManosMagnus:
Thank you @thc202 for the thorough review

PR #105, opened 2018-05-28
Add Nashorn load function example

Naramsim:


kingthorin:
@naramsim thanks for contributing, it ‘seems’ fine to me. Can you describe a use case?

thc202:
This is to reuse utility functions (like a library).

Ref: https://groups.google.com/d/topic/zaproxy-users/QlR4NLLqE1A/discussion

kingthorin:
Cool, thanks.

thc202:
The names of the scripts are too generic IMO, I'd suggest:
 - `import.js` → `load_function_example.js`
 - `export.js` → `example_library.js`

thc202:
Also worth adding a link to Nashorn docs.

standalone/load_function_example.js
```diff
@@ -0,0 +1,20 @@
+//@zaproxy-standalone
+
+// This script will load export.js and a file from the Internet
```
thc202: This comment needs to be updated to use the new file name.

standalone/load_function_example.js
```diff
@@ -0,0 +1,20 @@
...
+// Docs: https://wiki.openjdk.java.net/display/Nashorn/Nashorn+extensions
+
+print('loading scripts from: ' + java.lang.System.getProperty("user.dir"))
+
+var number = 0 // This variable will be overwritten by the loading of example_library.js
+
+// Load export.js
```
thc202: Same here.

thc202:
Thanks!

PR #106, opened 2018-06-07
HttpSender script to inject JS into an HTML page

Pamplemousse:
Inject inlined Javascript into the \<head\> tag of a targeted HTML page.

### DIFF
--- a/httpsender/inject_js_in_html_page.js
+++ b/httpsender/inject_js_in_html_page.js
@@ -22,7 +22,7 @@ function loadScriptFromFile(file) {
 function sendingRequest(msg, initiator, helper) {}
 
 function responseReceived(msg, initiator, helper) {
-    if (initiator == -1) { return; }
+    if (initiator != 1) { return; }
 
     body = msg.getResponseBody();
     bodyAsStr = body.toString();
--- a/httpsender/inject_js_in_html_page.js
+++ b/httpsender/inject_js_in_html_page.js
@@ -8,7 +8,6 @@
  */
 
 FILE = '/tmp/test.js'
-SCRIPT = '\t<script>' + loadScriptFromFile(FILE) + '</script>\n';
 
 function loadScriptFromFile(file) {
     Files = Java.type('java.nio.file.Files');
@@ -36,10 +35,19 @@ function responseReceived(msg, initiator, helper) {
         return true;
     }
 
-    index = body.toString().indexOf('<head>') + '<head>\n'.length();
+    SCRIPT = '<script>' + loadScriptFromFile(FILE) + '</script>';
+
+    indexOfHead = bodyAsStr.indexOf('<head>');
+
+    if (indexOfHead == -1) {
+      index = bodyAsStr.indexOf('<html>') + '<html>'.length();
+      SCRIPT = '<head>' + SCRIPT + '<head>';
+    } else {
+      index = indexOfHead + '<head>'.length();
+    }
 
     newBody = bodyAsStr.slice(0, index) + SCRIPT + bodyAsStr.slice(index);
-    header.setContentLength(newBody.length() + 1);
 
     msg.setResponseBody(newBody);
+    header.setContentLength(msg.getResponseBody().length());
 }
--- a/httpsender/inject_js_in_html_page.js
+++ b/httpsender/inject_js_in_html_page.js
@@ -1,0 +1,34 @@
+FILE = '/tmp/test.js'
+SCRIPT = '\t<script>' + loadScriptFromFile(FILE) + '</script>\n';
+
+function loadScriptFromFile(file) {
+    Files = Java.type('java.nio.file.Files');
+    Paths = Java.type('java.nio.file.Paths');
+    String = Java.type('java.lang.String');
+
+    filePath = Paths.get(file);
+    return new String(Files.readAllBytes(filePath), 'UTF-8');
+}
+
+function sendingRequest(msg, initiator, helper) {}
+
+function responseReceived(msg, initiator, helper) {
+    body = msg.getResponseBody();
+    bodyAsStr = body.toString();
+    header = msg.getResponseHeader();
+
+    xRequestedWith = msg.getRequestHeader().getHeader('X-Requested-With');
+    contentType = header.getHeader('Content-Type');
+
+    if (contentType != 'text/html; charset=utf-8'
+        || xRequestedWith == 'XMLHttpRequest') {
+        return true;
+    }
+
+    index = body.toString().indexOf('<head>') + '<head>\n'.length();
+
+    newBody = bodyAsStr.slice(0, index) + SCRIPT + bodyAsStr.slice(index);
+    header.setContentLength(newBody.length() + 1);
+
+    msg.setResponseBody(newBody);
+}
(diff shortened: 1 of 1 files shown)


thc202:
I guess using an extender script to add the listener would be better, it would avoid reading the file each time it needs to be injected. Also, might worth using a proxy listener (and inject after persisting to DB) to not let ZAP see the injection .

Pamplemousse:
I fear I do not really understand (can't find documentation) what extender scripts are for.

> ... it would avoid reading the file each time it needs to be injected.

It's kinda the point of this script: inject a JS file every time an html page is loaded through ZAP.
This file (containing JS) I can edit regularly without having to go through the cycle of developing a ZAP script.

As for the Proxy Listener, I do not know what they are.

thc202:
The script types are described in https://github.com/zaproxy/zap-core-help/wiki/HelpAddonsScriptsScripts (there's also some details in the READMEs of this repo) and in ZAP.
The extender script is installed/uninstalled into ZAP when enabled/disabled, so the functions/classes of the script are created just once instead of each time the script is invoked (as is the case for httpsender type), that way you can read the file you want to inject just once instead of each time, though that point is mute if you really want that. Worth pointing that out in the description.

>  inject a JS file every time an html page is loaded through ZAP.

That does not imply that you need to read it every time, you can read it once (or only when changed).

> I can edit regularly without having to go through the cycle of developing a ZAP script.

With the weekly the scripts can be edit outside ZAP, it will pick the new changes.

> As for the Proxy Listener, I do not know what they are.

A proxy listener (`org.parosproxy.paros.core.proxy.ProxyListener`) is like a proxy script but have more control on when the listener is called (so it can be called later after ZAP has persisted the message).

httpsender/inject_js_in_html_page.js
```diff
@@ -0,0 +1,45 @@
...
+ * Inject Javascript code into a webpage.
+ *
+ *   - Code will be added inside <script></script> tags in the Response body's
+ *   <head></head> before sending it to the client.
+ *   - Code to be added is read from the file which path is `FILE` (this need to
+ *   be an absolute path, /tmp/test.js in our case).
+ */
+
+FILE = '/tmp/test.js'
+SCRIPT = '\t<script>' + loadScriptFromFile(FILE) + '</script>\n';
```
thc202: Might worth doing this later, only when needed.

httpsender/inject_js_in_html_page.js
```diff
@@ -0,0 +1,45 @@
...
+    contentType = header.getHeader('Content-Type');
+
+    contentTypeRegex = new RegExp(/text\/html;/g);
+
+    if (!contentTypeRegex.test(contentType)
+        || xRequestedWith == 'XMLHttpRequest') {
+        return true;
+    }
+
+    index = body.toString().indexOf('<head>') + '<head>\n'.length();
```
thc202: - Could use `bodyAsStr`.
 - Ideally it should check that the `<head>` is present.
 - Should it include the `\n`? What if the file uses `\r\n` or does not have a newline?

httpsender/inject_js_in_html_page.js
```diff
@@ -0,0 +1,45 @@
...
+
+    newBody = bodyAsStr.slice(0, index) + SCRIPT + bodyAsStr.slice(index);
+    header.setContentLength(newBody.length() + 1);
```
thc202: Better use the length of the response body, which uses the byte count:
```JavaScript
msg.setResponseBody(newBody);
header.setContentLength(msg.getResponseBody().length());
```

httpsender/inject_js_in_html_page.js
Pamplemousse: Done.

httpsender/inject_js_in_html_page.js
Pamplemousse: Somewhat done.

Did not take care about line breaks: removed them.
Indeed you are right: there might be different policies, or even one-lined tags.

And breaking lines is not so important for injecting.
Use cases I target do not imply a need for good-looking html :)

httpsender/inject_js_in_html_page.js
Pamplemousse: Done.

httpsender/inject_js_in_html_page.js
```diff
@@ -0,0 +1,53 @@
...
+    header = msg.getResponseHeader();
+
+    xRequestedWith = msg.getRequestHeader().getHeader('X-Requested-With');
+    contentType = header.getHeader('Content-Type');
+
+    contentTypeRegex = new RegExp(/text\/html;/g);
+
+    if (!contentTypeRegex.test(contentType)
+        || xRequestedWith == 'XMLHttpRequest') {
+        return true;
```
thc202: This can be just `return;`, the function does not return anything and avoids confusions with proxy type (which does return boolean).

httpsender/inject_js_in_html_page.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    SCRIPT = '<script>' + loadScriptFromFile(FILE) + '</script>';
+
+    indexOfHead = bodyAsStr.indexOf('<head>');
+
+    if (indexOfHead == -1) {
+      index = bodyAsStr.indexOf('<html>') + '<html>'.length();
```
thc202: Most likely the html tag will have attributes, I'd just return if the head tag is not found (the script already says that the response body must have a head section so it would not be a surprise if it does not inject in that case).

httpsender/inject_js_in_html_page.js
```diff
@@ -0,0 +1,53 @@
...
+      SCRIPT = '<head>' + SCRIPT + '<head>';
```
thc202: It's not closing with the second head.

thc202:
Any thoughts on using a proxy script instead? Or, check the initiator for proxy requests in the current script? The httpsender is called for all messages sent by ZAP, injecting the script in active scanner responses does not sound right.

httpsender/inject_js_in_html_page.js
Pamplemousse: Done.

httpsender/inject_js_in_html_page.js
Pamplemousse: Done.

httpsender/inject_js_in_html_page.js
Pamplemousse: Removed this section as per previous comment.

thc202:
Could you add the initiator check (and tweak the description to mention the proxied responses)?

httpsender/inject_js_in_html_page.js
```diff
@@ -0,0 +1,48 @@
...
+    String = Java.type('java.lang.String');
+
+    filePath = Paths.get(file);
+    return new String(Files.readAllBytes(filePath), 'UTF-8');
+}
+
+function sendingRequest(msg, initiator, helper) {}
+
+function responseReceived(msg, initiator, helper) {
+    if (initiator == -1) { return; }
```
thc202: `!= 1`? https://github.com/zaproxy/community-scripts/tree/master/httpsender

thc202:
Thanks!

PR #127, opened 2018-10-04
Added additional authentication script & JavaScript examples

rezen:


psiinon:
Thanks for providing these - they look really useful!
It would be good to get them passing LGTM cleanly - hopefully thats not too much work?

psiinon:
This pull request **introduces 1 alert** when merging 750c2aa3245e0bfcf3c330ff419b883cf5cedc66 into b935afcd6d6f4153112f71b8b401928171fb8357 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/community-scripts/rev/pr-16b49f8e61762ebea146da6b2d8bd3937a37a318)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

kingthorin:
Rebased current, LGTM items fixed.

kingthorin:
I ended up tweaking this one. @psiinon @thc202 look good?

authentication/auth-api.js
```diff
@@ -0,0 +1,57 @@
+// Generic authentication script for authing against an API
+if (typeof println == 'undefined') this.println = print;
```
thc202: This workaround can be removed, `println` is not being used (ditto for the other scripts, the one that's using it can be changed to just use `print`). If we happen to need the workaround again we should implement it in core instead of requiring all scripts to add it.

httpsender/add-header-msg-id.js
```diff
@@ -0,0 +1,27 @@
...
+function responseReceived(msg, initiator, helper) {
+	var historyRef = msg.getHistoryRef();
+	var msgid = History.getLastHistoryId() + 1;
+
+	if (historyRef !== null) {
+		msgid = historyRef.getHistoryId();
+	}
+	
+	var header  = msg.getResponseHeader();
+	header.addHeader('X-Zap-Msgid', '' + msgid);
```
thc202: The ID sent is not be correct, the messages are persisted after being sent.

httpsender/greenbone-maintain-auth.js
```diff
@@ -0,0 +1,131 @@
...
+// Rewrite requests to include correct query token param
+function sendingRequest(msg, initiator, helper) {  
+  var reqbody = msg.getRequestBody().toString();
+  var headers = msg.getRequestHeader();
+  var url     = headers.getURI().toString();
+  var qry     = headers.getURI().getQuery();
+  var cookies = headers.getCookieParams();
+  
+  if (initiator === HttpSender.SPIDER_INITIATOR) {}
+  if (isStaticUrl(url)) {return true;}
```
thc202: The return values should be removed, the HTTP Sender scripts don't return anything.

httpsender/spider-wrangler.js
```diff
@@ -0,0 +1,119 @@
...
+  if (urlHasQuery) {return true;}
+  
+  logger("Checking for url without params " + url)
+
+  // If not in site tree ... move on
+  if (!isUrlInTree(url, rootNode, 0)) {return true;}
+
+  logger('Already have made this request ... dropping')
+  dropMessage(msg);
+  return false;
```
thc202: This has no effect, the message is still sent.

standalone/remove-x-status-codes.js
```diff
@@ -0,0 +1,62 @@
+// Remove 404s from request history
```
thc202: Should be updated to match the current behaviour (404 and 502s).

standalone/remove-x-status-codes.js
```diff
@@ -0,0 +1,62 @@
...
+var Control          = Java.type('org.parosproxy.paros.control.Control');
+var Model            = Java.type('org.parosproxy.paros.model.Model');
+
+var History  = Control.getSingleton().getExtensionLoader().getExtension(ExtensionHistory.class)
+var tree     = Model.getSingleton().getSession().getSiteTree()
+var rootNode = tree.getRoot();	
+
+function hasBadStatusCode(item) {
+  if (item === null) {return true;}
+  return item.getStatusCode() === 404 || item.getStatusCode() === 502 || item.getStatusCode() === null || item.getStatusCode() === 0;
```
thc202: It should not consider 0 status code as bad, that's for temporary Sites nodes, which should not be removed while in use.

targeted/test-trigger-script.js
```diff
@@ -0,0 +1,12 @@
+if (typeof println == 'undefined') this.println = print;
```
thc202: I'd suggest removing this script, not doing much more than the default.

thc202:
Some quick comments, still need to take a closer look.

httpsender/add-header-msg-id.js
thc202: Yeah, it might work in some simple cases, but if you have more than one request going out at the same time the IDs might no longer match (also worth noting that some of the messages might not be persisted at all).

httpsender/spider-wrangler.js
thc202: The Proxy scripts are able to drop the messages. I don't think there is.

rezen: Gotcha! So I should adjust the language from "drop" to something along the lines of "modifyToIgnore".

thc202: Sounds good.

rezen:
Re-submitting PR in smaller chunks

kingthorin:
@rezen are these still coming? I was going to point someone to your auth scripts then realized this was closed and they don't seem to have been added :cry:

rezen:
Sorry, yes! Added the PR!

PR #129, opened 2018-10-10
Delete HUNT - Debug & Logic Parameters.py

sbzo:
A new script was added optimizing time and processing. The new script is just one.


sbzo:
Merged into one PR.

PR #131, opened 2018-10-10
Delete HUNT - IDOR.py

sbzo:
A new script was added optimizing time and processing. The new script is just one View Pull Requests #128


sbzo:
Merged into one.

PR #142, opened 2018-12-13
[Find Emails.js] Improve E-Mail detection regex

nil0x42:
- ignores false positives such as '*@123' or '$@#!.'
- still detects e-mails with '+' portion
- checks domain portion correctness


thc202:
Thanks!

PR #143, opened 2018-12-21
find_reflected_params: refactor & optimize code

nil0x42:
refactoring & optimization:
    ---------------------------
    1. previous script was loading useless & greedy buffers
       reapeatedly, without being necessary.
       compare() function was called twice per check,
       and body & header buffers re-created lot of times.
    
    2. query parameters are now parsed using zaproxy built-in
       method getUrlParams() instead of reinventing the wheel.
    
    new features:
    -------------
    Add a `RESULT_PER_FINDING` boolean which lets user choose
    if he prefers to have results per reflected parameter,
    with evidence string properly hihglighter.
    (defaults to False)


passive/find_reflected_params.py
```diff
@@ -5,30 +5,51 @@
...
+            header = msg.getResponseHeader().toString()
+            body = msg.getResponseBody().toString()
+
+        if value in header or value in body:
+            if RESULT_PER_FINDING:
+                param_name = param.getName()
+                ps.raiseAlert(0, 2, alert_title, alert_desc, uri, param_name,
+                        None, None, None, value, 0, 0, msg)
+            else:
+                reflected_params.append(str(param.getName()))
```
thc202: The call to `str` is redundant.

passive/find_reflected_params.py
```diff
@@ -5,30 +5,51 @@
...
+    #               str param, str attack, str otherInfo, str solution,
+    # 		   str evidence, int cweId, int wascId, HttpMessage msg)
+    #  risk: 0: info, 1: low, 2: medium, 3: high
+    #  confidence: 0: falsePositive, 1: low, 2: medium, 3: high, 4: confirmed
+    alert_title = "Reflected HTTP GET parameter(s) (script)"
+    alert_desc = ("Reflected parameter value has been found. "
+                  "A reflected parameter values may introduce XSS "
+                  "vulnerability or HTTP header injection.")
+
+    # initialize empty vars
```
thc202: These comments could be removed, the code already says so.

passive/find_reflected_params.py
```diff
@@ -5,30 +5,51 @@
...
 Note that new passive scripts will initially be disabled
 Right click the script in the Scripts tree and select "enable"
-"""  
-def compare(paramvalue_pair,msg): 
-	reflected=''
-	if len(paramvalue_pair.split('=',1))>1:
-		value=paramvalue_pair.split('=',1)
-		body=msg.getResponseBody().toString()
-		header=msg.getResponseHeader().toString()
-		if (value[1] in body or value[1] in header) and value[1] != '' and len(value[1])>7:
```
thc202: The length check was removed, this will increase the FPs.

thc202:
Note that it would be fine to keep using this PR.

PR #144, opened 2018-12-21
 find_reflected_params: refactor & optimize code

nil0x42:
refactoring & optimization:
---------------------------
1. previous script was loading useless & greedy buffers
   reapeatedly, without being necessary.
   compare() function was called twice per check,
   and body & header buffers re-created lot of times.

2. query parameters are now parsed using zaproxy built-in
   method getUrlParams() instead of reinventing the wheel.

new features:
-------------
1. Add a `RESULT_PER_FINDING` boolean which lets user choose
   if he prefers to have results per reflected parameter,
   with evidence string properly hihglighter.
   (defaults to False)
2. Add a `MIN_PARAM_VALUE_LENGTH` int which lets user choose
   from which length parameter values will be checked
   (defaults to 8)

### DIFF
--- a/passive/find_reflected_params.py
+++ b/passive/find_reflected_params.py
@@ -7,28 +7,28 @@ Note that new passive scripts will initially be disabled
 Right click the script in the Scripts tree and select "enable"
 """  
 def compare(paramvalue_pair,msg): 
-	reflected=''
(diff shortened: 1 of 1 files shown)


thc202:
Thanks!

PR #165, opened 2019-08-06
Add download concat script

fabianfrz:
copied changes from zaproxy/zap-extensions#2137 to here.


fabianfrz:
Just in case: I fixed your unit test which is broken in Java 11 for all ruby files.

The exclusion forgets to remove the file from the list.

thc202:
There were no Ruby scripts in this repo yet, that's why it went unnoticed... Thanks for fixing it!

PR #179, opened 2020-03-24
Ported from SQLMap - apostrophenullencode

sbzo:


thc202:
Cherry picked into #180.

PR #184, opened 2020-06-01
Merge FuzzerStopOnStatusCode200 script

jtommi:
I added a httpfuzzerprocessor script.
I tried to make it work with a parameter, but the first use parameter would not get overwritten by the ones entered for further fuzzes

### DIFF
--- a/httpfuzzerprocessor/FuzzerStopOnStatusCode200
+++ b/httpfuzzerprocessor/FuzzerStopOnStatusCode200
@@ -1,0 +1,25 @@
+/**
+ * Processes the fuzzed message (payloads already injected).
+ * 
+ * Called before forwarding the message to the server.
+ * 
+ * @param {HttpFuzzerTaskProcessorUtils} utils - A utility object that contains functions that ease common tasks.
+ * @param {HttpMessage} message - The fuzzed message, that will be forward to the server.
+ */
+function processMessage(utils, message) {
+}
+
+/**
+ * Processes the fuzz result.
+ * 
+ * Called after receiving the fuzzed message from the server.
+ * 
+ * @param {HttpFuzzerTaskProcessorUtils} utils - A utility object that contains functions that ease common tasks.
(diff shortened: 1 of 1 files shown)


thc202:
The script needs the extension (`.js`).
The JSDoc could be removed, that's provided in the template and other places.

thc202:
> but the first use parameter would not get overwritten by the ones entered for further fuzzes

Could you provide the steps to reproduce that?

jtommi:
> 
> 
> The script needs the extension (`.js`).
> The JSDoc could be removed, that's provided in the template and other places.

I thought to myself: let's quickly add the file through the github webgui... That's how you forget file extensions.



> 
> 
> > but the first use parameter would not get overwritten by the ones entered for further fuzzes
> 
> Could you provide the steps to reproduce that?

Sure:
1. Choose something to fuzz
1. Right-click and Fuzz
1. Add a payload
1. Go to message processors and click "Add"
1. Choose the script
1. Enter a status code in the parameter field (which doesn't appear with the committed script because I removed it)
1. Click Add and Start Fuzzer
1. Start all over but this time enter a different status code
1. You should "see" that the fuzzer stops as soon as it gets the initial status code, not the second one

If you can help, I'd love to make that work because it would make the script way more versatile

thc202:
Thanks, I'm not able to reproduce that:
```JavaScript
var STATUS_CODE_PARAM = "Status Code"

function processMessage(utils, message) {
}

function processResult(utils, fuzzResult){
	var code = utils.getParameters().get(STATUS_CODE_PARAM)
	print(code)
	return true
}

function getRequiredParamsNames(){
	return [STATUS_CODE_PARAM]
}
```
prints always the latest status code set in the UI. Could you give it another try or share your script using parameters?

jtommi:
I guess I did something wrong, unfortunately I didn't keep the old code.
With your implementation, it works fine.
Thanks

PR #186, opened 2020-06-23
Fix typo in HTTP Message Logger

thc202:
Correct the name of the constant `Integer.MAX_VALUE`, which caused an
error in other ECMAScript engine than Nashorn.


psiinon:
All good but now has conflicts

thc202:
Addressed.

PR #218, opened 2021-04-06
Add active scan script for Cross-Site WebSocket Hijacking validation

pf-msi:
Signed-off-by: Piotr Furman <piotr.furman@motorolasolutions.com>

### DIFF
--- a/active/Cross Site WebSocket Hijacking.js
+++ b/active/Cross Site WebSocket Hijacking.js
@@ -1,0 +1,109 @@
+/**
+ * Copyright (C) 2021 Motorola Solutions, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Author: Piotr Furman <piotr.furman at motorolasolutions.com>
+ *
+ * Description:
+ *
+ * Script to validate potential Cross-Site WebSocket Hijacking vulnerability.
+ *
+ * According to RFC 6455, section 10.2 Origin Considerations:
+ *
+ * "Servers that are not intended to process input from any web page but
+ * only for certain sites SHOULD verify the |Origin| field is an origin
+ * they expect.  If the origin indicated is unacceptable to the server,
+ * then it SHOULD respond to the WebSocket handshake with a reply
+ * containing HTTP 403 Forbidden status code."
+ *
+ * Which means we can try to repeat WebSocket HTTP Upgrade request with
+ * a modified Origin header and raise an alert if it was accepted.
+ *
+ * Note: Run ajax spider before a scan in order to evaluate your application's
+ *       JavaScript code which opens WebSocket connection.
+ * Note: Active scripts are initially disabled, right click the script to enable it.
+ */
+
+var Base64 = Java.type("java.util.Base64")
+var Random = Java.type("java.util.Random")
+var String = Java.type("java.lang.String")
+var ByteArray = Java.type("byte[]")
+
+var LOG_DEBUG_MESSAGES = false // change to true for more logs
+
+var RISK = 3
+var CONFIDENCE = 2
+var TITLE = "Cross-Site WebSocket Hijacking"
+var DESCRIPTION = "Server accepted WebSocket connection through HTTP Upgrade request with modified Origin header."
+var SOLUTION = "Validate Origin header on WebSocket connection handshake, to ensure only specified origins are allowed to connect.\
+ Also, WebSocket handshake should use random tokens, similar to anti CSRF tokens."
+var REFERENCE = "https://tools.ietf.org/html/rfc6455#section-10.2"
+var OTHER = "See also https://portswigger.net/web-security/websockets/cross-site-websocket-hijacking\
+ or https://christian-schneider.net/CrossSiteWebSocketHijacking.html"
+var CWEID = 346 // CWE-346: Origin Validation Error, http://cwe.mitre.org/data/definitions/346.html
+var WASCID = 9 // WASC-9 Cross Site Request Forgery, http://projects.webappsec.org/w/page/13246919/Cross%20Site%20Request%20Forgery
+
+function scanNode(as, msg) {
+    var target = msg.getRequestHeader().getURI().toString()
+
+    // check if this is a WebSocket HTTP Upgrade request (the message should include also "Connection: Upgrade" header if we wanted to check it strictly)
+    // TODO: in ZAP 2.11 we might use msg.isWebSocketUpgrade() check instead
+    var upgradeHeader = msg.getRequestHeader().getHeader("Upgrade")
+    if (!upgradeHeader || upgradeHeader.toLowerCase() !== "websocket") {
+        if (LOG_DEBUG_MESSAGES) {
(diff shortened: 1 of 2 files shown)


active/Cross Site WebSocket Hijacking.js
```diff
@@ -0,0 +1,100 @@
...
+var CWEID = 346 // CWE-346: Origin Validation Error, http://cwe.mitre.org/data/definitions/346.html
+var WASCID = 9 // WASC-9 Cross Site Request Forgery, http://projects.webappsec.org/w/page/13246919/Cross%20Site%20Request%20Forgery
+
+function scanNode(as, msg) {
+    var target = msg.getRequestHeader().getURI().toString()
+
+    // check if this is a WebSocket HTTP Upgrade request
+    // request message should include also "Connection: Upgrade" header but let's skip it for now, and switch to msg.isWebSocketUpgrade() if https://github.com/zaproxy/zaproxy/pull/6508 was merged
+    var upgrade_header = msg.getRequestHeader().getHeader("Upgrade")
+    if (!upgrade_header || upgrade_header.toLowerCase() !== "websocket") {
```
kingthorin, line 57: Should the second part be a contains check instead of not equal? (Just to cover a greater set of circumstances)

active/Cross Site WebSocket Hijacking.js
```diff
@@ -0,0 +1,100 @@
...
+        print("Cross-Site WebSocket Hijacking scan skipped for url=" + target + ", is not a WebSocket upgrade request")
```
kingthorin, line 58:
```suggestion
        print("Cross-Site WebSocket Hijacking scan skipped for url=" + target + ", it does not appear to be a WebSocket upgrade request.")
```

active/Cross Site WebSocket Hijacking.js
```diff
@@ -0,0 +1,100 @@
...
+        return
+    }
+
+    print("Cross-Site WebSocket Hijacking scan started for url=" + target)
```
kingthorin, line 62:
```suggestion
    print("Cross-Site WebSocket Hijacking rule started for url=" + target)
```

active/Cross Site WebSocket Hijacking.js
thc202, line 58: These messages should be wrapped under a flag (default to false), otherwise this will generate a lot of noise.

active/Cross Site WebSocket Hijacking.js
```diff
@@ -0,0 +1,100 @@
...
+    new Random().nextBytes(randomBytes)
+    var secWsKey = new String(Base64.getEncoder().encode(randomBytes))
+    msg.getRequestHeader().setHeader("Sec-WebSocket-Key", secWsKey)
+
+    // set Origin header using custom domain, .example is a reserved TLD in RFC 2606 so it should not match domain name of a scanned service
+    msg.getRequestHeader().setHeader("Origin", "https://cswsh.example")
+
+    as.sendAndReceive(msg, true, false)
+
+    if (as.isStop()) {
```
thc202, line 76: This can be removed, or if you think it's really needed do before sending the message. At this point it might as well do the quick status code check.

active/Cross Site WebSocket Hijacking.js
```diff
@@ -0,0 +1,100 @@
...
+    if (responseStatus == 101) {
+        // should not have accepted connection with different origin
+        print("Cross-Site WebSocket Hijacking vulnerability found, sending alert for url=" + target)
+        as.newAlert()
+          .setRisk(RISK)
+          .setConfidence(CONFIDENCE)
+          .setName(TITLE)
+          .setDescription(DESCRIPTION)
+          .setParam(target)
+          .setAttack('')
```
thc202, line 90: Remove.

active/Cross Site WebSocket Hijacking.js
```diff
@@ -0,0 +1,100 @@
...
+var REFERENCE = "https://tools.ietf.org/html/rfc6455#section-10.2"
+var OTHER = "See also https://portswigger.net/web-security/websockets/cross-site-websocket-hijacking or https://christian-schneider.net/CrossSiteWebSocketHijacking.html"
+var CWEID = 346 // CWE-346: Origin Validation Error, http://cwe.mitre.org/data/definitions/346.html
+var WASCID = 9 // WASC-9 Cross Site Request Forgery, http://projects.webappsec.org/w/page/13246919/Cross%20Site%20Request%20Forgery
+
+function scanNode(as, msg) {
+    var target = msg.getRequestHeader().getURI().toString()
+
+    // check if this is a WebSocket HTTP Upgrade request
+    // request message should include also "Connection: Upgrade" header but let's skip it for now, and switch to msg.isWebSocketUpgrade() if https://github.com/zaproxy/zaproxy/pull/6508 was merged
```
thc202, line 55: Assume that it will be merged instead, and make this a TODO.

active/Cross Site WebSocket Hijacking.js
```diff
@@ -0,0 +1,100 @@
+/**
```
thc202, line 1: This mixing of copyright/license with other information is confusing. Maybe split into two comments?

active/Cross Site WebSocket Hijacking.js
pf-msi: For WebSocket handshakes "upgrade" header should be equal to "websocket", it won't have any other value. According to specification, handshake should have also a `connection: upgrade` header. And you're right, "connection" might contain multiple entries. 

But in practice, it should be enough if we check just the `Upgrade: websocket` header. AFAIK there's no scenario when the browser would send proper "upgrade" header, and "connection" header without "upgrade" value in the same request. 

And once we switch to `msg.isWebSocketUpgrade()`, proper check will be executed in there.

active/Cross Site WebSocket Hijacking.js
pf-msi: Thanks, fixed.

active/Cross Site WebSocket Hijacking.js
pf-msi: Thanks, done. Is there a way to check current ZAP logging level from inside the script?

active/Cross Site WebSocket Hijacking.js
pf-msi: Thanks, done.

active/Cross Site WebSocket Hijacking.js
pf-msi: Ok, done.

active/Cross Site WebSocket Hijacking.js
pf-msi: Ok, I reordered license and description.

kingthorin: I think his point was more that they should be completely separate blocks.

active/Cross Site WebSocket Hijacking.js
kingthorin: Sounds good, thanks

active/Cross Site WebSocket Hijacking.js
pf-msi: @thc202 is this comment issue blocking this pull request? I reordered comment text, so the copy and license is not mixed with script's description. Of course I can split it into two separated comment blocks if it makes a difference.

kingthorin: @thc202 ^ ?

active/Cross Site WebSocket Hijacking.js
thc202: Sure, using the Log4j APIs.

active/Cross Site WebSocket Hijacking.js
thc202: It would be better to split them, easier to read.

thc202:
I'm going to address the remaining comments and fix the changelog.

thc202:
Rebased to correct the changelog.

thc202:
Comments addressed.

thc202:
Removed empty line added accidentally.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thank you both!

pf-msi:
Thanks for completing the PR!

PR #233, opened 2021-06-25
Remove not so useful passive scripts

Prince-Mendiratta:
Signed-off-by: Prince Mendiratta <prince.mendiratta@getastra.com>


Prince-Mendiratta:
Apologies, a mistake.

PR #240, opened 2021-08-18
Add unexpected_responses.js script

rbailey-godaddy:
This is another variation on the "do something if we see a surprising response" theme, but is intended to be more flexible:

* The script uses parameters rather than hard-coded values
* Response codes are tested against a regular expression rather than a literal value
* A sense code is used to indicate if matching the regex is "good" or "bad", so you can either whitelist or blacklist depending on your particular use case
* Failures (unexpected responses) generate alerts. The formatting here was tailored for mechanical consumption, but there are suggestions in the comments if you value "pretty" output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - active/Cross Site WebSocket Hijacking.js > an active scan for Cross-Site WebSocket Hijacking vulnerability
 - targeted/cve-2021-22214.js > A targeted script to check for Unauthorised SSRF on GitLab - CVE 2021-22214
 - httpsender/full-session-n-csrf-nashorn.js > full session and csrf token management.
+- httpfuzzerprocessor/unexpected_responses.js > compare response codes to regex whitelist/blacklist and generate alerts
 
 ### Changed
 - Update links in READMEs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - active/Cross Site WebSocket Hijacking.js > an active scan for Cross-Site WebSocket Hijacking vulnerability
 - targeted/cve-2021-22214.js > A targeted script to check for Unauthorised SSRF on GitLab - CVE 2021-22214
 - httpsender/full-session-n-csrf-nashorn.js > full session and csrf token management.
-- httpfuzzerprocessor/unexpected_responses.js > compare response codes to regex whitelist/blacklist and generate alerts
+- httpfuzzerprocessor/unexpected_responses.js > compare response codes to a (pass/fail) regex and generate alerts
 
 ### Changed
 - Update links in READMEs.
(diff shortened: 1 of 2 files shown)


kingthorin:
Could you update https://github.com/zaproxy/community-scripts/blob/main/CHANGELOG.md with an added note in the unreleased section?

Also you seem to have committed with two different accounts, if that's something that matters to you, feel free to tweak it and force push.

rbailey-godaddy:
> Could you update https://github.com/zaproxy/community-scripts/blob/main/CHANGELOG.md with an added note in the unreleased section?

Done.

> Also you seem to have committed with two different accounts, if that's something that matters to you, feel free to tweak it and force push.

I don't care that much -- they're both me. ;-)

kingthorin (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - active/Cross Site WebSocket Hijacking.js > an active scan for Cross-Site WebSocket Hijacking vulnerability
 - targeted/cve-2021-22214.js > A targeted script to check for Unauthorised SSRF on GitLab - CVE 2021-22214
 - httpsender/full-session-n-csrf-nashorn.js > full session and csrf token management.
+- httpfuzzerprocessor/unexpected_responses.js > compare response codes to regex whitelist/blacklist and generate alerts
```
thc202, line 11:
```suggestion
- httpfuzzerprocessor/unexpected_responses.js > compare response codes to (a pass/fail) regex and generate alerts
```

httpfuzzerprocessor/unexpected_responses.js
```diff
@@ -0,0 +1,137 @@
+/*
+ * This fuzz processor script will raise alerts based on HTTP response codes.
+ * It expects two special parameters:
+ * - A regular expression used to match the response code
+ * - A string that is either "pass" or "fail", indicating whether or not a
+ *   matching response is expected ("pass") or unexpected ("fail").
+ * Unexpected responses will cause an alert to be raised.
+ */
+
+// See https://github.com/zaproxy/community-scripts/tree/master/httpfuzzerprocessor
```
thc202, line 10:
```suggestion
// See https://github.com/zaproxy/community-scripts/tree/main/httpfuzzerprocessor
```

httpfuzzerprocessor/unexpected_responses.js
```diff
@@ -0,0 +1,137 @@
...
+
+// See https://github.com/zaproxy/community-scripts/blob/master/httpfuzzerprocessor/showDifferences.js
```
thc202, line 12:
```suggestion
// See https://github.com/zaproxy/community-scripts/blob/main/httpfuzzerprocessor/showDifferences.js
```

CHANGELOG.md
rbailey-godaddy: I made this edit very slightly differently "...a (pass/fail) regex..."

httpfuzzerprocessor/unexpected_responses.js
rbailey-godaddy: Done

httpfuzzerprocessor/unexpected_responses.js
rbailey-godaddy: Done

thc202 (approved):

thc202:
Thank you!

PR #242, opened 2021-09-02
New add-more-headers.js httpsender script

rbailey-godaddy:
A list of headers specified in "add_headers" global variable will be added to all requests.

Signed-off-by: Scott Bailey <scott.bailey@godaddy.com>

### DIFF
--- a/httpsender/add-more-headers.js
+++ b/httpsender/add-more-headers.js
@@ -1,0 +1,70 @@
+// This HttpSender script adds headers to all messages transmitted by zaproxy,
+// including automated tools. Refer to the HttpSender class definition:
+// https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
+// for a list of 'initiator' values (although we don't use them).
+
+var ScriptVars = Java.type("org.zaproxy.zap.extension.script.ScriptVars");
+
+/*
+ * HttpSender scripts do not support parameters, so we'll use a known global
+ * variable to supply desired content. The value of this variable should be a
+ * JSON string containing a serialized map<String, String> object. The map keys
+ * are the desired header name and the values are the header values.
+ *
+ * Example:
+ * add_headers defined with value '{"x-this": "v1", "x-that": "v2"}' will
+ * result in the following headers being added to every request:
+ * x-this: v1
+ * x-that: v2
+ */
+
+PARAMETER_VARIABLE = "add_headers";
+user_headers = null;
+
+// Logging with the script name is super helpful!
+function logger() {
+	print('[' + this['zap.script.name'] + '] ' + arguments[0]);
+}
+
+// Parse and store headers where we can get at them quickly
+function initializeHeaders(variableName) {
+	logger("Initializing...");
+	user_headers = JSON.parse(ScriptVars.getGlobalVar(variableName));
+}
+
+/*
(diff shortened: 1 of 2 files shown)


httpsender/add-more-headers.js
```diff
@@ -0,0 +1,64 @@
...
+// https://github.com/zaproxy/zaproxy/blob/master/zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
+// for a list of 'initiator' values (although we don't use them).
+
+var ScriptVars = Java.type("org.zaproxy.zap.extension.script.ScriptVars");
+
+/*
+ * HttpSender scripts do not support parameters, so we'll use a known global
+ * variable to supply desired content. The value of this variable should be a
+ * JSON string containing a serialized map<String, String> object. The map keys
+ * are the desired header name and the values are the header values.
```
kingthorin, line 12: Including an example would probably be a good move. I think I see what you're meaning but I'm not sure others will.
```json
{"X-Forward-For":"localhost","some-other-header":"some-value"}
```

httpsender/add-more-headers.js
```diff
@@ -0,0 +1,64 @@
...
+
+// Logging with the script name is super helpful!
+function logger() {
+	print('[' + this['zap.script.name'] + '] ' + arguments[0]);
+}
+
+// Parse and store headers where we can get at them quickly
+function initializeHeaders(variableName) {
+	logger("Initializing...");
+	user_headers = JSON.parse(ScriptVars.getGlobalVar(variableName));
```
thc202, line 26: How is the global variable being set? Wondering if this could use custom variable (thus get an `Object` instead of a `String`).

rbailey-godaddy: We're doing it like this (from python) at present:

```python
    # Store the result in a global variable where it can be accessed later
    header_json = json.dumps(header_dict)
    log.debug("Setting user headers: %s", header_json)
    zap.script.set_global_var("add_headers", header_json)
```

The `zap` object is instantiated by `zapv2.ZAPv2()`.

thc202: OK, from the API needs to be as is then.

httpsender/add-more-headers.js
rbailey-godaddy: I added an example (and found a statement that was misleading).

kingthorin (approved):
LGTM

thc202:
Could you address the conflict?

thc202 (approved):

thc202:
Conflict addressed.

Thank you!

httpsender/add-more-headers.js
```diff
@@ -0,0 +1,70 @@
+// This HttpSender script adds headers to all messages transmitted by zaproxy,
+// including automated tools. Refer to the HttpSender class definition:
+// https://github.com/zaproxy/zaproxy/blob/master/zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```
thc202, line 3: Needs update.

thc202: Done.

rbailey-godaddy: Thanks, just circled back and you're ahead of me :)

thc202: No problem, rushing for release :)

kingthorin (approved):

PR #253, opened 2021-10-09
fix: cve-2021-41773-apache-path-trav.js set path escaped

kingthorin:
- Tweak targeted script, set the path as escaped so that things are further encoded making it fail. Remove undefined pluginid, use -1.
- Add changelog note.

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
-
+### Fixed
+- targeted/cve-2021-41773-apache-path-trav.js - Set path as escaped so that it's handled properly, set pluginid properly.
 
 ## [12] - 2021-10-07
 ### Added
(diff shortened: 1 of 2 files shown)


kingthorin:
Tested against a dockerized vuln apache `docker run -dit --name CVE-2021-41773 -p 8080:80 -v /opt/apache2.4.49:/usr/local/apache2/htdocs httpd:2.4.49`

thc202 (approved):

kingthorin:
Ok, that makes more sense. Thank you!

kingthorin:
Ref: https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/nashorn/api.html look for Java.extend

thc202:
You need to call `setURI()` after the `_path` assignment.

kingthorin:
`<eval>:104 ReferenceError: "setUri" is not defined<eval>:104 ReferenceError: "setUri" is not defined` or if I use `this.setUri()` I get `<eval>:104 TypeError: this.setUri is not a function<eval>:104 TypeError: this.setUri is not a function` :cry:
`Java.super(newUri).setUri();` also results in "not a function"

PR #279, opened 2022-05-31
Update js scripts to use passed variables

ricekot:


thc202 (approved):

thc202:
Thank you!

kingthorin:
As discussed earlier do other script languages need similar handling and could/should we be handilng at the add-on or engine vs in each script?

thc202:
This handling is temporary, once 2.12 is out the checks will no longer be needed.

kingthorin:
Perfect, thanks.

PR #280, opened 2022-06-29
Check for the version in Server Header Disclosure 

prateekrana-getastra:
The Alert Description states about leaking version information via the http header. The previous version for the script only checks for the header without considering whether even the version is leaked or not. Added checks for looking up numeric versions information available in the http header.

### DIFF
--- a/passive/Server Header Disclosure.js
+++ b/passive/Server Header Disclosure.js
@@ -1,7 +1,9 @@
 // Server Header Check by freakyclown@gmail.com
+// Server Version leaks found via header field by prateek.rana@getastra.com
 
-function scan(ps, msg, src) 
-{
+var VERSION_PATTERN = new RegExp("(?:\\d+\\.)+\\d+");
+
+function scan(ps, msg, src)  {
 
     var alertRisk = 1
     var alertConfidence = 2
@@ -15,9 +17,24 @@ function scan(ps, msg, src)
     var url = msg.getRequestHeader().getURI().toString();
     var headers = msg.getResponseHeader().getHeaders("Server")
     
-    if (headers != null)
+    if (headers != null && containsPotentialSemver(headers))
     {
-        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution,headers, cweId, wascId, msg);
+        var headersString = headers.toString();
+        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution, headersString, cweId, wascId, msg);
     }
     
+}
+
+function containsPotentialSemver(content) {
+    try {
+        var res = VERSION_PATTERN.exec(content);
+        if (res == null || res.join('') === ""){
+            return false;
+        }
+        return true;
+    }
+
+    catch (err) {
+        return false;
+    }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - standalone/enableDebugLogging.js > Updated for more recent logging funtionality.
 - Update JS scripts to use passed singleton variables (control, model, view) if available (>= ZAP 2.12.0).
+- passive/Server Header Disclosure.js > Updated to check that the Server Header contains something that looks like a semantic version component.
 
 ## [14] - 2021-11-01
 ### Added


passive/Server Header Disclosure.js
```diff
@@ -14,10 +15,38 @@ function scan(ps, msg, src)
...
     {
-        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution,headers, cweId, wascId, msg);
+        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution, headers_string, cweId, wascId, msg);
     }
     
 }
+
+function ExtAlert(content) {
+
+    var ext = new RegExp("(\\d+\\.)+\\d+");
```
kingthorin, line 29: Why does this have a capturing group if the potential matches aren't returned?

You could make it non-capturing with `?:` .
https://www.stefanjudis.com/today-i-learned/non-capturing-groups-in-regular-expressions/

kingthorin (changes requested):

passive/Server Header Disclosure.js
```diff
@@ -14,10 +15,38 @@ function scan(ps, msg, src)
 
     var url = msg.getRequestHeader().getURI().toString();
     var headers = msg.getResponseHeader().getHeaders("Server")
+    var headers_string = headers.toString();
```
thc202, line 18: This will fail if the `Server` header is not present.

passive/Server Header Disclosure.js
```diff
@@ -14,10 +15,38 @@ function scan(ps, msg, src)
...
+            return false;
+        }
+        res = res.join('');
+
+        if (res === "") {
+            return false;
+        }
+        
+        else {
+            print("Server version leak found via header.");
```
thc202, line 43: Remove, there's already an alert raised.

passive/Server Header Disclosure.js
```diff
@@ -14,10 +15,38 @@ function scan(ps, msg, src)
...
+            return true;        
+        }
+    }
+
+    catch (err) {
+        print(err);
```
thc202, line 49: This is no that useful as is, include an informative message or just remove it.

thc202:

passive/Server Header Disclosure.js
```diff
@@ -14,10 +15,38 @@ function scan(ps, msg, src)
...
-    if (headers != null)
+    if (headers != null && ExtAlert(headers))
     {
-        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution,headers, cweId, wascId, msg);
+        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution, headers_string, cweId, wascId, msg);
     }
     
 }
+
+function ExtAlert(content) {
```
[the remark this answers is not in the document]
prateekrana-getastra: made the change.

passive/Server Header Disclosure.js
```diff
@@ -14,10 +15,38 @@ function scan(ps, msg, src)
...
+    try {
+        var res = ext.exec(content);
+        if (res == null){
+            return false;
+        }
+        res = res.join('');
+
+        if (res === "") {
+            return false;
+        }
```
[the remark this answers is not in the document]
prateekrana-getastra: combined them.

passive/Server Header Disclosure.js
prateekrana-getastra: used non-capture group.

passive/Server Header Disclosure.js
prateekrana-getastra: added this inside function after checking for header != null

passive/Server Header Disclosure.js
prateekrana-getastra: removed

passive/Server Header Disclosure.js
prateekrana-getastra: removed.

passive/Server Header Disclosure.js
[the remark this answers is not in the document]
kingthorin: It looks like you didn't do this quite properly.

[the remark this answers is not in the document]
kingthorin: As LGTM pointed out the join is now dangling and unreachable (it should have been moved into the if as outlined above), the else isn't necessary. It's returns false at the if, if that condition isn't hit then the code can continue on and just return true (without the else).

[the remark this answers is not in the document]
prateekrana-getastra: Yes. I understood it differently. Fixed it now.

passive/Server Header Disclosure.js
```diff
@@ -15,9 +16,31 @@ function scan(ps, msg, src)
...
+
+        else {
+            return true;        
+        }
+
+    }
+
+    catch (err) {
+        return false;
+    }
```
kingthorin, line 45: As previously mentioned the else isn't necessary.

```js
function containsPotentialServer(content) {

    var ext = new RegExp("(?:\\d+\\.)+\\d+");

    try {
        var res = ext.exec(content);
        if (res == null || res.join('') === "") {
            return false;
        }
        return true;        
     }
    catch (err) {
        return false;
    }
```

prateekrana-getastra: Removed it.

prateekrana-getastra:
> The changelog should be updated.

Changes to be added under Unreleased?

kingthorin:
Yes, for this one: Under Unreleased, in the `### Changed` section.

passive/Server Header Disclosure.js
```diff
@@ -15,9 +16,27 @@ function scan(ps, msg, src)
...
+    if (headers != null && containsPotentialServer(headers))
     {
-        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution,headers, cweId, wascId, msg);
+        var headers_string = headers.toString();
+        ps.raiseAlert(alertRisk, alertConfidence, alertTitle, alertDesc, url, '', '', '', alertSolution, headers_string, cweId, wascId, msg);
     }
     
 }
+
+function containsPotentialServer(content) {
```
kingthorin, line 27: `containsPotentialSemver` I didn't notice this earlier, sorry about that.

I had meant for it to purposefully be "semver" meaning short form for 'Semantic Version' (ex: https://semver.org)

prateekrana-getastra: Should be as intended now.

prateekrana-getastra:
Anything else to update?

kingthorin:
⚠️ Please don't make any further changes without resetting your local copy first:
```bash
git fetch origin
git reset --hard origin/version-leak-found
```

thc202 (approved):

thc202:
Thank you both!

PR #284, opened 2022-10-02
Update build/workflows, compile with Java 11

thc202:
Update Gradle to 7.5.1, compile the add-on with Java 11.
Update workflows' actions and Java version.


kingthorin (approved):

.github/workflows/ci.yml
```diff
@@ -11,21 +11,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [8, 11]
+        java: [11]
```
ricekot, line 14: Should we add Java 17 as well?

thc202: I didn't add to this repo because it almost has no Java code (only a single simple class), can add if you think it's a must.

ricekot (approved):
Thanks!

PR #328, opened 2023-04-12
Update ZAP to 2.12

thc202:
Remove compatibility code that's no longer needed, that provided singletons.
Use provided singletons (`control` and `model`) in Python scripts.
Use non-deprecated `HttpSender` constructor and use constant instead of the literal for the initiator value (clearer).


psiinon (approved):

kingthorin (approved):

kingthorin:
Thanks!

PR #343, opened 2023-05-19
Lots of regex passive scripts

kurobeats:


kingthorin (approved):
There's probably a load of places that groups should be non-capturing, but that's a minor detail I think.

thc202:

PR #354, opened 2023-06-23
JWT Decode Script

0mgfriday:
Adds a script for decoding JWTs.

image

Addresses https://github.com/zaproxy/zaproxy/issues/7654


encode-decode/JwtDecode.js
```diff
@@ -0,0 +1,42 @@
+// JWT Decode by 0mgfriday
+var Base64 = Java.type("java.util.Base64")
+var String = Java.type("java.lang.String")
+
+/**
+ * Decode JWT into JSON
```
thc202, line 6: It does not decode into JSON, "into a text representation" would be more accurate.

encode-decode/JwtDecode.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    return helper.newError("Invalid JWT");
+}
+
+function b64decode(s) {
+    s = s.replace('-', '+')
+    s = s.replace('_', '/')
+    var bytes = Base64.getDecoder().decode(s)
```
thc202, line 36: Use the URL variant instead of manual replacements.

encode-decode/JwtDecode.js
```diff
@@ -0,0 +1,42 @@
...
+    return new String(bytes)
```
thc202, line 37: Better use a specific charset.

encode-decode/JwtDecode.js
0mgfriday: updated

encode-decode/JwtDecode.js
0mgfriday: Updated

kingthorin:
To address the DCO requirement you'll need to sign-off the commit(s):

- https://github.com/zaproxy/zaproxy/blob/main/CONTRIBUTING.md#developer-certificate-of-origin
- https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff

You may as well also squash or fix-up the commits.

encode-decode/JwtDecode.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Decode JWT into a text representation
+ *
+ * @param {EncodeDecodeScriptHelper} helper - A helper object with various utility methods.
+ *     For more details see https://github.com/zaproxy/zap-extensions/tree/main/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
+ * @param {String} value - JWT to decode
+ * @returns {EncodeDecodeResult} - Decoded JWT (JSON)
+ */
+function process(helper, value){
+    const parts = value.split('.')
```
thc202, line 15: Use `var`  (and for the `result`), to make it work with both Nashorn and Graal.

thc202:
The commit message could be tidied up.

thc202 (approved):

psiinon (approved):

PR #362, opened 2023-07-03
Create ParamDiggerDelayScan.js

Vicvio:
This script will delay requests sent using the param digger according to a user-defined value in milliseconds, in cases where there are strict requirements about using automatic tools on a Web application during the security assessment

### DIFF
--- a/httpsender/Param Digger Delay Scan
+++ b/httpsender/Param Digger Delay Scan
@@ -1,0 +1,36 @@
+// This script introduces a time delay between requests sent using the Param-Digger Extension
+// You may change it to apply to other Initiators such as the traditional Spider
+// Initiators such as the Active Scanner and Fuzzer come with built-in functionality to set time delays
+
+
+
+// 'initiator' is the component that initiated the request:
+//  17 PARAM_DIGGER_INITIATOR in this case
+// For the latest list of values see the HttpSender class:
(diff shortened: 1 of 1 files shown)


httpsender/Param Digger Delay Scan
```diff
@@ -0,0 +1,36 @@
...
+	// leave the response alone
+}
+
+// create time delay between requests
+function wait(ms) {
+    var start = Date.now(),
+        now = start;
+    while (now - start < ms) {
+      now = Date.now();
+    }
```
thc202, line 35: This wastes cycles, use `Thread.sleep`.

thc202:
Please see https://github.com/zaproxy/community-scripts/blob/main/CONTRIBUTING.md

It worth noting that a more general way of rate limiting will be added through the Network add-on.

httpsender/Param Digger Delay Scan
Vicvio: Nice point. Will change it... thanks

thc202:
Closing per the available [global rate limiting](https://www.zaproxy.org/docs/desktop/addons/network/options/ratelimit/) feature, while rate limiting with HTTP Sender script seems useful it does not actually allow to enforce the limits (as the user can easily/accidentally exceed them, e.g. active scan, manual).

PR #374, opened 2023-08-02
other: Relocate some content

kingthorin:
Move "api", "af-plans", and "scan-hooks" into "other".


kingthorin:
Note I'm happy to include a CHANGELOG message but wasn't sure if it was necessary since these items aren't really ZAP scripts for users to use within ZAP.

thc202 (approved):

thc202:
I don't think it's necessary to mention in the changelog.

Thank you!

ricekot (approved):
Thanks!

PR #429, opened 2024-01-13
Add HTTP Sender script to RSA sign the requests

mwalkowski:
Hi All,

I've prepared a Python script for ZAP that adds an RSA digital signature based on a private key. I encountered such a solution in one of my projects, and I think it's worth sharing. I've also prepared a similar solution for Burp and described everything under this [link](https://mwalkowski.github.io/post/using-burp-python-scripts-to-sign-requests-with-rsa-keys/).


thc202:

httpsender/rsa-signing-for-zap.py
thc202, line 1: Ideally the name should be camel case https://github.com/zaproxy/community-scripts/blob/main/CONTRIBUTING.md#naming-scripts

mwalkowski: Corrected

mwalkowski:
> The changelog should be updated.

I updated changelog

psiinon (approved):

thc202:
Thank you!

PR #438, opened 2024-03-07
Add MFA handling script in Selenium

njmulsqb:
I am testing an application that has MFA implemented so I wrote this script to handle MFA in auth flow. I have verified its working with authenticated requestor job in AF.

Feedback on improvements shall be appreciated!

With 💚 from Tecvity


kingthorin (commented):
Please update the CHANGELOG. Otherwise looks good to me.

kingthorin (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   to find and add subdomains to the Sites Tree.
 - passive/JavaDisclosure.js - Passive scan for Java error messages leaks
 - httpsender/RsaEncryptPayloadForZap.py - A script that encrypts requests using RSA
+- selenium/Fill OTP in MFA.js - A script that fills the OTP in MFA
```
thc202, line 12: Please take a look at https://github.com/zaproxy/community-scripts/blob/main/CONTRIBUTING.md

njmulsqb: Ok I'll fix the name; Actually I took inspiration from the naming convention followed in the scripts already present in repo i.e. https://github.com/zaproxy/community-scripts/blob/main/selenium/Selenium%20Juice%20Shop.js

selenium/Fill OTP in MFA.js
```diff
@@ -0,0 +1,21 @@
+// This script can be used to fill in the OTP if it appears right after the authentication.
+// This can only work for the apps who have fixed OTP in MFA
```
psiinon, line 2: I think more context would be really useful here.
For example, in this case you have configured browser based auth - this works but then the MFA prompt appears.
It would really help to have a snippet of the HTML of the MFA dialog including the relevant bits like 'flt-glass-pane' and 'flt-semantics-placeholder' - this will help people adapt it to their situations.
This technique will also work if its possible to get the MFA token locally, eg via another command line tool. You dont have to implement that, just maybe mention it?

njmulsqb: The `flt-glass-pane` thing is only applicable to Flutter apps, I guess we can remove this line as this will probably be useless for all the audience that is not using flutter app and just adds to the complexity of script. What do you say?

Other than that, for HTML do you want me to add a snippet containing OTP input field and submit button?

psiinon: Yep, sounds good. Its just to help people who are have a similar situation but which will obviously be slightly diffferent..

CHANGELOG.md
kingthorin: Ya we will have to crack down on that 🙂

njmulsqb: @kingthorin May I create an issue for that in main repo?

thc202: https://github.com/zaproxy/community-scripts/pull/438#discussion_r1516227144

Not sure there's the need to rename old scripts, which would break users' workflows.

https://github.com/zaproxy/community-scripts/pull/438#discussion_r1516229631

The issue would be in this repo: https://github.com/zaproxy/community-scripts/issues

psiinon (approved):

Issue #170, opened 2019-10-24
Normalise and enforce script formatting/indentation


thc202:
Per title, might be asking too much given the number of languages...

sourabhdeshmukh:
@thc202 could you please give more details regarding which scripts needs formating.

thc202:
The ones under the script type directories (e.g. [`active`](https://github.com/zaproxy/community-scripts/tree/master/active)).

Visionvxyz:
pull下载研究研究

kingthorin:
I started to look at this based on spotless. I figured first I should look at implementing it for JavaScript. Using the [eclipse web tool platform](https://github.com/diffplug/spotless/tree/main/plugin-gradle#eclipse-web-tools-platform) formatter fails (it does run but bombs with syntax errors). 

<details>
<summary>Stack & Console output</summary>

```console
$ ./gradlew spotlessApply

> Task :spotlessJs FAILED
Skipping 'C:\Users\thorin\Desktop\zap-ws\community-scripts\active\Cross Site WebSocket Hijacking.js' because it does not converge.  Run `spotlessDiagnose` to understand why
Skipping 'C:\Users\thorin\Desktop\zap-ws\community-scripts\active\JWT None Exploit.js' because it does not converge.  Run `spotlessDiagnose` to understand why
Skipping 'C:\Users\thorin\Desktop\zap-ws\community-scripts\active\User defined attacks.js' because it does not converge.  Run `spotlessDiagnose` to understand why
Step 'eclipse wtp formatters - JS' found problem in 'extender\HTTP Message Logger.js':
Invalid JavaScript syntax for formatting.
java.lang.IllegalArgumentException: Invalid JavaScript syntax for formatting.
        at com.diffplug.spotless.extra.eclipse.wtp.EclipseJsFormatterStepImpl.format(EclipseJsFormatterStepImpl.java:96)
        at jdk.internal.reflect.GeneratedMethodAccessor21.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at com.diffplug.spotless.extra.wtp.EclipseWtpFormatterStep.lambda$applyWithoutFile$1(EclipseWtpFormatterStep.java:64)
        at com.diffplug.spotless.FormatterFunc.apply(FormatterFunc.java:32)
[...]

PR #441, opened 2024-04-09
Format JS scripts with Prettier

ricekot:
Format all JS scripts with Prettier via the Spotless Gradle plugin.

The dependency on `npm` to run prettier is fulfilled via the [gradle-node-plugin](https://github.com/node-gradle/gradle-node-plugin).

Part of #170.


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Cool

Issue #440, opened 2024-04-04
Add ScanRuleMetadata to Active and Passive Scan Rules


ricekot:
Tracker issue for adding [ScanRuleMetadata](https://github.com/zaproxy/zaproxy/issues/7105) to active and passive scripts in this repo.

There are some scripts in the list below which may clash with existing rules in one of the scan rule add-ons (e.g. the check done by `X-Powered-By_header_checker.js` script is also done by the [XPoweredByHeaderInfoLeakScanRule](https://github.com/zaproxy/zap-extensions/blob/f292e25b523bc29f2d3e1339dd14d5891e8efede/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/XPoweredByHeaderInfoLeakScanRule.java) scan rule). We should either remove the script or the scan rule to avoid duplicate alerts.

Also, WebSocket passive scripts are not included here since they don't support the `getMetadata()` function at the moment.

## Active

- [ ] bxss.py
- [ ] corsair.py
- [x] Cross Site WebSocket Hijacking.js
- [x] cve-2019-5418.js
- [x] gof_lite.js (Duplicate of: [BackupFileDisclosureScanRule.java](https://github.com/zaproxy/zap-extensions/blob/678f11f46c8a4542d3eccb469cb519d079b5802c/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosureScanRule.java))
- [x] JWT None Exploit.js
- [ ] RCE.py
- [x] SSTI.js
- [ ] SSTI.py
- [ ] TestInsecureHTTPVerbs.py
- [ ] User defined attacks.js

## Passive

- [x] clacks.js
- [x] CookieHTTPOnly.js
- [x] detect_csp_notif_and_reportonly.js
- [x] detect_samesite_protection.js
- [x] f5_bigip_cookie_internal_ip.js
- [x] find base64 strings.js
- [x] Find Credit Cards.js
- [x] Find Emails.js
- [x] Find Hashes.js
- [x] Find HTML Comments.js
- [x] Find IBANs.js
- [x] Find Internal IPs.js
- [x] find_reflected_params.py
- [x] google_api_keys_finder.js
- [x] HUNT.py
- [x] JavaDisclosure.js
- [x] Mutliple Security Header Check.js
- [x] Report non static sites.js
- [x] RPO.js
- [x] s3.js
- [x] Server Header Disclosure.js
[...]

PR #442, opened 2024-04-14
Implement `getMetadata` for active JS scripts

ricekot:
Part of #440.

Also add relevant alert tags and assign categories. The scan rule IDs for these scripts were taken from the [scanners.md](https://github.com/zaproxy/zaproxy/blob/aedad6818c562a741115333a660d5de09f1780d9/docs/scanners.md) file.

The status of these scripts has been set to `alpha` since community-scripts is an alpha add-on.

### DIFF
--- a/active/Cross Site WebSocket Hijacking.js
+++ b/active/Cross Site WebSocket Hijacking.js
@@ -41,31 +41,45 @@ var Base64 = Java.type("java.util.Base64");
 var Random = Java.type("java.util.Random");
 var String = Java.type("java.lang.String");
 var ByteArray = Java.type("byte[]");
+var ScanRuleMetadata = Java.type(
+  "org.zaproxy.addon.commonlib.scanrules.ScanRuleMetadata"
+);
+var CommonAlertTag = Java.type("org.zaproxy.addon.commonlib.CommonAlertTag");
 
 var LOG_DEBUG_MESSAGES = false; // change to true for more logs
 
-var RISK = 3;
-var CONFIDENCE = 2;
-var TITLE = "Cross-Site WebSocket Hijacking";
-var DESCRIPTION =
-  "Server accepted WebSocket connection through HTTP Upgrade request with modified Origin header.";
-var SOLUTION =
-  "Validate Origin header on WebSocket connection handshake, to ensure only specified origins are allowed to connect.\
- Also, WebSocket handshake should use random tokens, similar to anti CSRF tokens.";
-var REFERENCE = "https://tools.ietf.org/html/rfc6455#section-10.2";
-var OTHER =
-  "See also https://portswigger.net/web-security/websockets/cross-site-websocket-hijacking\
- or https://christian-schneider.net/CrossSiteWebSocketHijacking.html";
-var CWEID = 346; // CWE-346: Origin Validation Error, http://cwe.mitre.org/data/definitions/346.html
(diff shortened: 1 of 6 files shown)


kingthorin:
FYI I haven't forgotten this. I did go through a number and they seemed fine. I'll try to finish a review in the next few days.

psiinon:
image

Looks pretty good 😁 
Can we add meta data so we can link to the script source code? This might require code changes of course.
If it does then maybe also change to be able to report the alert type?
It would be better if it was "Alert Script" in this case

kingthorin:
I finally got to read through all of them, seems fine to me. Are there further changes/tweaks coming or shall I go ahead and approve?

kingthorin (approved):
Thanks!

active/Cross Site WebSocket Hijacking.js
```diff
@@ -41,31 +41,44 @@ var Base64 = Java.type("java.util.Base64");
...
+  - https://tools.ietf.org/html/rfc6455#section-10.2
+category: server
+risk: high
+confidence: medium
+cweId: 346  # CWE-346: Origin Validation Error, http://cwe.mitre.org/data/definitions/346.html
+wascId: 9  # WASC-9 Cross Site Request Forgery, http://projects.webappsec.org/w/page/13246919/Cross%20Site%20Request%20Forgery
+alertTags:
+  OWASP_2021_A01: https://owasp.org/Top10/A01_2021-Broken_Access_Control/
+  OWASP_2017_A05: https://owasp.org/www-project-top-ten/2017/A5_2017-Broken_Access_Control.html
+  WSTG-v42-CLNT-10: https://owasp.org/www-project-web-security-testing-guide/v42/4-Web_Application_Security_Testing/11-Client-side_Testing/10-Testing_WebSockets
```
thc202, line 68: We should look at a way of using the common lib for these as well.

active/JWT None Exploit.js
```diff
@@ -4,18 +4,35 @@ var Cookie = Java.type("java.net.HttpCookie");
...
+  which bypasses the JWT hash verification.
+solution: >
+  Use a secure JWT library, and (if your library supports it) restrict the allowed hash algorithms.
+references:
+  - https://www.sjoerdlangkemper.nl/2016/09/28/attacking-jwt-authentication/
+category: server
+risk: high
+confidence: medium
+cweId: 347
+wascId: 15
```
thc202, line 26: Comments were dropped but in the other script they were kept (don't mind either way just saying for consistency).

active/JWT None Exploit.js
```diff
@@ -128,20 +145,9 @@ function scanNode(as, msg) {
...
-    SOLUTION,
-    "Cookie: " + cookie.getName() + "=" + payload,
-    REFERENCE,
-    CWEID,
-    WASCID,
-    msg
-  );
+  as.newAlert()
+    .setEvidence("Cookie: " + cookie.getName() + "=" + payload)
+    .setUri(msg.getRequestHeader().getURI().toString())
```
thc202, line 150: Can be removed, it's the default from the message set.

active/gof_lite.js
```diff
@@ -139,41 +145,16 @@ function mutate(msg, mutationString) {
 
 function raiseAlert(wasSuccess, msg, origMsg, as) {
   var url = msg.getRequestHeader().getURI().toString();
```
thc202, line 147: The url could be dropped.

thc202 (approved):

thc202:
Thank you!

thc202:
@psiinon do you want to check this again?

psiinon (approved):

PR #443, opened 2024-05-03
tips: Add Edge as Chrome

kingthorin:
- Relevant tip content and images.


kingthorin:
I should add a note that while this might be handy for producing it probably isn't a good move for the Ajax Spider or DOM XSS rule.

kingthorin:
Probably something like:

```md
 > [!CAUTION] 
> While this tip might be handy for exploring an app manually it probably IS NOT a good move to set Edge as Chrome for the purposes of the Ajax Spider or DOM XSS rule.
```

kingthorin:
Tweaked, ready for review.

ricekot (approved):
Thanks!

Is it worth adding the tips to the website too?

kingthorin:
Good question. We originally did it here in the hopes that community would contribute, but that hasn't really taken off.

other/tips/selenium/edge/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+![](images/edge_properties.png)
+
+5. Go back to ZAP and into the Selenium options:
+6. Set the Chrome binary location to point to the Edge executable path you copied earlier.
+
+![](images/selenium_options.png)
+
+7. Then when you launch a 'Chrome' browser (either via the toolbar button or manual explore screen), it will launch Edge.
+8. Note: You may get a message saying the "Provided browser was not found" when you close Edge :shrug: (just click "Ok" and ignore it.)
```
thc202, line 29: Did you try use with the official WebDriver for Edge?
https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/

kingthorin: And specifying that as the chrome driver? No I didn't.

I can/will 😃

kingthorin: Unfortunately in that case (switching chrome driver for msedgedriver) it refuses to launch the browser at all 😞

kingthorin: Same if I rename the driver :wink:

thc202: Ok, they do have their own `EdgeDriver` class which should address that. The error is not the best but... we could also look at adding support for that (even if we don't bundle the WebDriver).

thc202 (approved):

thc202:
Thank you!

PR #447, opened 2024-05-31
feat: add Ory Kratos authentication scripts

getlarge:
## Description

I added two authentication scripts to enable authenticated interacted with web applications relying on [Ory Kratos](https://github.com/ory/kratos).
- `KratosApiAuthentication.js` adds support for the [Native Login flow](https://www.ory.sh/docs/kratos/self-service/flows/user-login#login-for-api-clients-and-clients-without-browsers)
- `KratosABrowserAuthentication.js` adds support for the [Browser Login flow](https://www.ory.sh/docs/kratos/self-service/flows/user-login#login-for-server-side-browser-clients)

I hope they may be useful for others in the Zap community.

### DIFF
--- a/authentication/KratosApiAuthentication.js
+++ b/authentication/KratosApiAuthentication.js
@@ -1,8 +1,10 @@
 /*
  * This script is used to authenticate a user using Ory Kratos self-service API for clients without browsers.
  *
- * Authentication Verification can be configured with a Logged In Regex as something like:
- *  \Qactive.*
+ * Authentication Verification can be configured with the following rule:
+ * - Verification Strategy: Poll the Specified URL
+ * - Regex Pattern used to identify Logged In message: \Qactive.*
+ * - URL to poll: <Kratos Base URL>/sessions/whoami
  *
  * Zap must be configured to use HTTP header-based session management with the value: {%json:session_token%}
  *
--- a/authentication/KratosApiAuthentication.js
+++ b/authentication/KratosApiAuthentication.js
@@ -11,12 +11,12 @@
  */
 
 const HttpRequestHeader = Java.type(
-  'org.parosproxy.paros.network.HttpRequestHeader'
+  "org.parosproxy.paros.network.HttpRequestHeader"
 );
-const HttpHeader = Java.type('org.parosproxy.paros.network.HttpHeader');
-const URI = Java.type('org.apache.commons.httpclient.URI');
+const HttpHeader = Java.type("org.parosproxy.paros.network.HttpHeader");
+const URI = Java.type("org.apache.commons.httpclient.URI");
 const AuthenticationHelper = Java.type(
-  'org.zaproxy.zap.authentication.AuthenticationHelper'
+  "org.zaproxy.zap.authentication.AuthenticationHelper"
 );
 
 /**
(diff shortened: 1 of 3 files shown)


thc202:
The commits need to be signed-off: https://github.com/zaproxy/community-scripts/pull/447/checks?check_run_id=25640107047
You need to run the `spotlessApply` task (e.g. `./gradlew sA`) to fix the formatting of the scripts. 
The changelog should be updated to mention the new scripts: https://github.com/zaproxy/community-scripts/blob/main/CONTRIBUTING.md#changelog

getlarge:
> The commits need to be signed-off: https://github.com/zaproxy/community-scripts/pull/447/checks?check_run_id=25640107047 You need to run the `spotlessApply` task (e.g. `./gradlew sA`) to fix the formatting of the scripts. The changelog should be updated to mention the new scripts: https://github.com/zaproxy/community-scripts/blob/main/CONTRIBUTING.md#changelog

Sorry, I missed the contribution guidelines. Too bad I had to remove my GPG signature ;)
Should be good now!

getlarge:
Yes, but not during the rebase unfortunately.

getlarge:
@thc202 Can you help for the reviewing of this PR? Is anything blocking?

authentication/KratosApiAuthentication.js
```diff
@@ -0,0 +1,135 @@
+/*
+ * This script is used to authenticate a user using Ory Kratos self-service API for clients without browsers.
+ *
+ * Authentication Verification can be configured with the following rule:
+ * - Verification Strategy: Poll the Specified URL
+ * - Regex Pattern used to identify Logged In message: \Qactive.*
```
thc202, line 6: This pattern does not look right, missing `\E` before `.*`? Or remove `\Q`? (Some of the comments added in this script also apply to the other.)

authentication/KratosApiAuthentication.js
```diff
@@ -0,0 +1,135 @@
...
+ * - URL to poll: <Kratos Base URL>/sessions/whoami
+ *
+ * Zap must be configured to use HTTP header-based session management with the value: {%json:session_token%}
```
thc202, line 9: ZAP

authentication/KratosApiAuthentication.js
```diff
@@ -0,0 +1,135 @@
...
+ * @returns {Object} The HTTP message used to perform the authentication.
+ */
+function authenticate(helper, paramsValues, credentials) {
+  print("Authenticating via Ory Kratos...");
+
+  // Step 1: Initialize the login flow
+  const kratosBaseUri = paramsValues.get("Kratos Base URL");
+  const initLoginUri = new URI(
+    kratosBaseUri + "/self-service/login/api",
+    false
```
thc202, line 54: Would be better if this was true, having the URI already escaped/encoded avoids surprises.

getlarge: Honestly here I just took inspiration from the other Authentication snippets which all seems to set false as a value for the second parameter.
If I understand correctly, you would suggest to do something like:
```js
new URI(encodeURI(kratosBaseUri + "/self-service/login/api"), true);
```
for all URI instantiation?

thc202: Yes.

authentication/KratosApiAuthentication.js
getlarge: Oh yes, thanks for pointing this out! That makes the tests much faster ;)

getlarge:
@thc202 Just applied your suggestions.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #450, opened 2024-06-07
export_site-tree_URLs.js

a6k8s4:
This script can export all the Spidered/Scanned URLs from the site tree for a web-application/site while automation job is on the go. The script can work on both GUI & CLI with integration with ZAP automation framework addon.

### DIFF
--- a/standalone/export_site-tree_URLs.js
+++ b/standalone/export_site-tree_URLs.js
@@ -1,0 +1,55 @@
+// Standalone scripts have no template.
+// They are only evaluated when you run them.
+//By Ankush
+//This script can export all the spidered/scanned urls from the site tree while automation job is on the go. 
+var ZAP_API = 'http://localhost:8080';
+var API_KEY = 'q0bhhpgs2oc18k0o';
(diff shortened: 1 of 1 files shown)


psiinon:
I'm confused - why use the API when you can access everything directly?
See https://github.com/zaproxy/community-scripts/blob/main/standalone/Traverse%20sites%20tree.js

a6k8s4:
> I'm confused - why use the API when you can access everything directly? See https://github.com/zaproxy/community-scripts/blob/main/standalone/Traverse%20sites%20tree.js

Hi Simon, the purpose of using ZAP API is , the format of output data are the better than the traverse sites tree script. The format is better one for logging as per the necessity. This is pretty helpful while the automation job is running from CLI.
For example.
image

thc202:
You can output any way you want while traversing the tree directly.

thc202:
Could you update the script to use the tree directly?

a6k8s4:
> Could you update the script to use the tree directly?

Sure.. I will update. Right now I am using the API for getting the desired result..

kingthorin:
@a6k8s4 😉

PR #452, opened 2024-06-12
Implement `getMetadata` for some more Passive scripts

ricekot:
Update the following scripts to implement the `getMetadata()` function:
- passive/google_api_keys_finder.js
- passive/JavaDisclosure.js
- passive/Report non static sites.js
- passive/RPO.js
- passive/s3.js

Part of #440.

### DIFF
--- a/passive/Report non static sites.js
+++ b/passive/Report non static sites.js
@@ -4,58 +4,61 @@
 // Note that new passive scripts will initially be disabled
 // Right click the script in the Scripts tree and select "enable"
 
+var ScanRuleMetadata = Java.type(
+  "org.zaproxy.addon.commonlib.scanrules.ScanRuleMetadata"
+);
+
+function getMetadata() {
+  return ScanRuleMetadata.fromYaml(`
+id: 100017
+name: Non Static Site Detected
+description: >
+  A query string or form has been detected in the HTTP response body.
+  This indicates that this may not be a static site.
+solution: >
+  If this is not a static site then ignore or disable this rule.
+risk: info
+confidence: medium
+status: alpha
+codeLink: https://github.com/zaproxy/community-scripts/blob/main/passive/Report%20non%20static%20sites.js
+helpLink: https://www.zaproxy.org/docs/desktop/addons/community-scripts/
+`);
+}
+
 /**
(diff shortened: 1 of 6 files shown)


passive/Report non static sites.js
```diff
@@ -4,58 +4,61 @@
...
+function getMetadata() {
+  return ScanRuleMetadata.fromYaml(`
+id: 100017
+name: Non Static Site Detected
+description: >
+  A query string or form has been detected in the HTTP response body.
+  This indicates that this may not be a static site.
+solution: >
+  If this is not a static site then ignore or disable this rule.
+risk: high
```
kingthorin, line 20: I understand you've recreated what was previously here but ummm imho we are safe to change this

kingthorin: I guess if it is changed that should be noted in the changelog

ricekot: Should I change it to `info`?

kingthorin: That seems reasonable to me

passive/s3.js
```diff
@@ -1,22 +1,29 @@
...
-    "Remove s3 Buckets name from response or make sure the permissions in bucket are configured properly.";
-  var cweId = 200;
-  var wascId = 13;
+var ScanRuleMetadata = Java.type(
+  "org.zaproxy.addon.commonlib.scanrules.ScanRuleMetadata"
+);
 
+function getMetadata() {
+  return ScanRuleMetadata.fromYaml(`
+id: 100035
```
thc202, line 9: Same ID as `Information Disclosure - Java Stack Trace` (also in `scanners.md`).

passive/Report non static sites.js
thc202: Regarding the changelog, some names/descriptions/… were changed as well (not only in this PR), maybe add a generic entry saying that the scripts that now implement `getMetadata()` had their info revised?

passive/s3.js
ricekot: Wow, funny how we all missed that.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #490, opened 2025-09-06
add find-js-file.js to passive rules

nmwafa:
new passive scan script to find js file

### DIFF
--- a/passive/find-js-file.js
+++ b/passive/find-js-file.js
@@ -1,0 +1,60 @@
+// Author: https://nmwafa.github.io - with GPT
+
+var ScanRuleMetadata = Java.type(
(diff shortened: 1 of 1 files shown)


psiinon:
Is this really a helpful rule?
In my experience apps can reference a _lot_ of JS files. ZAP should find them via spidering so they will be in the Sites Tree. I'm not sure we need Info alerts for them as well.

`Review all referenced JavaScript files. Ensure they do not contain 
  sensitive data (e.g., API keys, credentials) and follow secure coding practices.`

JS files can be _huge_ so I'm not sure this advice is particularly helpful. It would be much better to implement a rule which actually checked for the sensitive data automatically 😁

